Issue #49, opened 2022-01-11
Test Retry Logic


tustvold:
Currently there is no real testing of the retry logic, in particular leader failover, connection errors, etc... This should be addressed

PR #74, opened 2022-01-19
refactor: Add a little testability, abstract a bit over broker kinds

carols10cents:
Connects to #49, in that it adds a tiny bit of testing around getting a new arbitrary broker connection to cache.

This also adds a bit of type info around different sources of broker connection, and fixes a subtle bug where broker id was being logged as `None` even though we had a broker ID from the topology.

Again recommended to review commit-by-commit.

I feel less confident about the value of these extractions and the new test. If the new test fails, it's likely to be a flaky failure too because of the randomness.

Looking at other code like the controller and partition leader caching, I think perhaps the broker caching trait could be reused, and then perhaps the `maybe_retry` functions could be shared more and tested more? Not sure yet though.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -44,7 +44,7 @@ pub type Result<T, E = Error> = std::result::Result<T, E>;
 trait ConnectionHandler {
     type R: RequestHandler + Send + Sync;
 
-    async fn connection(
+    async fn connect(
         &self,
         tls_config: TlsConfig,
         socks5_proxy: Option<String>,
@@ -81,7 +81,7 @@ impl BrokerRepresentation {
 impl ConnectionHandler for BrokerRepresentation {
     type R = MessengerTransport;
 
-    async fn connection(
+    async fn connect(
         &self,
         tls_config: TlsConfig,
         socks5_proxy: Option<String>,
@@ -199,7 +199,7 @@ impl BrokerConnector {
         match self.topology.get_broker(broker_id).await {
             Some(broker) => {
                 let connection = BrokerRepresentation::Topology(broker)
-                    .connection(
+                    .connect(
                         self.tls_config.clone(),
                         self.socks5_proxy.clone(),
                         self.max_message_size,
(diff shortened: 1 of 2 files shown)


src/connection.rs
```diff
@@ -482,4 +569,60 @@ mod tests {
     fn arbitrary_recoverable_error() -> RequestError {
         RequestError::IO(std::io::Error::from(std::io::ErrorKind::UnexpectedEof))
     }
+
+    struct FakeBrokerConn {
+        conn: Box<dyn Fn() -> Result<Arc<FakeConn>> + Send + Sync>,
+    }
+
+    #[derive(Debug, PartialEq)]
+    struct FakeConn;
```
crepererum, line 578: The names confuse me a bit. What's the difference between a "fake broker conn(ection)" and a "fake conn(ection)"?

src/connection.rs
```diff
@@ -37,14 +37,79 @@ pub enum Error {
 
 pub type Result<T, E = Error> = std::result::Result<T, E>;
 
+/// How to connect to a `Transport`
```
crepererum, line 40: If you connect to a "transport", then why is `async fn connection` returning `R: Request`? Is this a bit of a naming clash because `Transport` was already taken by `crate::connection::transport::Transport`? Maybe we can call `Request` a `RequestHandler`. I think nouns are better for traits and structs and verbs rather for functions. Same goes for `Connect::connection` which I would rather call `ConnectionHandler::connect`, esp. since the method actually establishes a new connection rather than using a cached one (we currently don't cache connections at this specific layer and I think we don't want to because Kafka doesn't really multiplex TCP connection requests but only allows in-order pipelining).

crepererum (approved):
The interface / abstraction point is good, I'm a bit confused about naming things (which IIRC is the 2nd hard problem of computer science).

This conflicts a bit w/ #72 but it shouldn't be too hard of either of us to rebase (depending on who goes first) since it's effectively only affecting the arguments that `Connct::connection` is going to pass through to `Transport`.

src/connection.rs
carols10cents: Yeaaahhh I didn't put a whole lot of thought into the names 😅 I renamed `FakeBrokerConn` to `FakeBrokerRepresentation`, is that better?

src/connection.rs
carols10cents: Yup, these names were not great. I've taken your suggestions here, how does it look now?

carols10cents:
Because you approved, I'm going to merge this. If we need to adjust names more, I'm happy to do so in a future PR!

PR #75, opened 2022-01-20
test: fix remaining protocol bits and fix remaining issues

### LINKED ISSUE #64: Protocol fuzzing
Write a fuzzer for our binary read protocol. This should randomly select an ApiKey and ApiVersion and provide a byte vector from which we'll try to read a response header followed by the response body. This must not panic.

I'm pretty sure this is going to panic btw. (I have seen this happen during fiddling with the protocol) because while we ensure that the message size is capped, we don't have any size constraints when reading variable-sized collections (like strings and arrays) from the message stream. E.g. this is going to panic for large arrays/strings sizes even though the message size is small:

https://github.com/influxdata/rskafka/blob/377b1c104fa05bec5a45b92ddea3815cd507c8b8/src/protocol/primitives.rs#L569

https://github.com/influxdata/rskafka/blob/377b1c104fa05bec5a45b92ddea3815cd507c8b8/src/protocol/primitives.rs#L373

We should probably cap our allocations.

crepererum:
Closes #64.

### DIFF
--- a/src/protocol/vec_builder.rs
+++ b/src/protocol/vec_builder.rs
@@ -76,7 +76,7 @@ impl VecBuilder<u8> {
         R: std::io::Read,
     {
         // Note: We can modify `self` here and still return an error because there is no way the taken/moved `self` will
-        //       be accessible in the error case.S
+        //       be accessible in the error case.
 
         while self.remaining_elements > 0 {
             let mut buf = self.blocks.last_mut().expect("Has always at least 1 block");
--- a/src/protocol/vec_builder.rs
+++ b/src/protocol/vec_builder.rs
@@ -71,10 +71,13 @@ impl<T> VecBuilder<T> {
 
 impl VecBuilder<u8> {
     /// Read as many bytes as there are left.
-    pub fn read_exact<R>(&mut self, reader: &mut R) -> Result<(), std::io::Error>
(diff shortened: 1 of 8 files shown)


src/protocol/vec_builder.rs
```diff
@@ -71,10 +71,13 @@ impl<T> VecBuilder<T> {
 
 impl VecBuilder<u8> {
     /// Read as many bytes as there are left.
-    pub fn read_exact<R>(&mut self, reader: &mut R) -> Result<(), std::io::Error>
+    pub fn read_exact<R>(mut self, reader: &mut R) -> Result<Self, std::io::Error>
     where
         R: std::io::Read,
     {
+        // Note: We can modify `self` here and still return an error because there is no way the taken/moved `self` will
+        //       be accessible in the error case.S
```
carols10cents, line 79: Typo? 

```suggestion
        //       be accessible in the error case.
```

carols10cents (approved):
Looks pretty good with the limited experience I have, Raphael should probably review too.

Do you think it's worth auditing all non-test uses of:

- ` as `
- `vec![_, N]`
- `Vec::with_capacity`

crepererum:
I've audited the two vector cases already. I'll file another PR for the `as` conversions and just delegate it to clippy.

tustvold (approved):

PR #77, opened 2022-01-21
feat: allow `Aggregator::flush` to return an error

crepererum:
This simplifies the integration w/ certain clients like IOx and there's
no real reason why the other methods can error and this one cannot.

Also adds testing for the entire aggregator error handling which was
previously missing.


src/client/producer.rs
```diff
@@ -307,12 +307,18 @@ where
     status_deagg: <A as aggregator::Aggregator>::StatusDeaggregator,
 }
 
+#[derive(Debug, Clone)]
+enum AggregatedError {
+    Aggregator(Arc<aggregator::Error>),
```
tustvold: I think this is an Arc of a box which seems a tad redundant?

src/client/producer.rs
```diff
@@ -307,12 +307,18 @@ where
...
+    Client(Arc<ClientError>),
+}
+
 #[derive(Debug)]
 struct AggregatedResult<A>
 where
     A: aggregator::Aggregator,
 {
-    inner: Result<Arc<AggregatedStatus<A>>, Arc<ClientError>>,
+    inner: Result<Arc<AggregatedStatus<A>>, AggregatedError>,
```
tustvold: I personally would prefer `Arc<AggregatedError>` over a cloneable error enumeration

src/client/producer.rs
```diff
@@ -213,7 +213,7 @@ mod broadcast;
 #[derive(Debug, Error)]
 pub enum Error {
     #[error("Aggregator error: {0}")]
-    Aggregator(#[from] aggregator::Error),
+    Aggregator(#[from] Arc<aggregator::Error>),
```
tustvold, line 216: Same as below, we have `Arc` of `Box` here

src/client/producer.rs
```diff
@@ -307,12 +307,18 @@ where
     status_deagg: <A as aggregator::Aggregator>::StatusDeaggregator,
 }
 
+#[derive(Debug, Clone)]
+enum AggregatedError {
```
tustvold: This seems to just be a subset of `Error`?

src/client/producer.rs
```diff
@@ -435,7 +435,18 @@ where
...
         trace!("Flushing batch producer");
 
-        let (output, status_deagg) = inner.aggregator.flush();
+        let (output, status_deagg) = match inner.aggregator.flush() {
+            Ok(x) => x,
+            Err(e) => {
+                // Reset result slot
+                let slot = std::mem::take(&mut inner.result_slot);
+
+                slot.broadcast(AggregatedResult {
```
tustvold, line 444: It's personal preference and I know this PR didn't introduce it, but perhaps `AggregatedResult` could just be a type alias instead of a custom struct. This would then make this just `slot.broadcast(AggregatedError::Aggregator(Arc::new(e)))`

tustvold (approved):

src/client/producer.rs
```diff
@@ -210,10 +210,10 @@ use crate::record::Record;
 pub mod aggregator;
 mod broadcast;
 
-#[derive(Debug, Error)]
+#[derive(Debug, Error, Clone)]
```
tustvold, line 213: Comment got orphaned by another change, but I wonder if using `Arc<Error>` vs a clone-able error might be cleaner? :thinking:

src/client/producer.rs
crepererum: done

src/client/producer.rs
crepererum: done

src/client/producer.rs
crepererum: aggregated error is gone

src/client/producer.rs
crepererum: fixed

src/client/producer.rs
crepererum: I found this easier to adapt from the users point of view since it's basically an error pass-through.

PR #79, opened 2022-01-21
WIP more refactoring

carols10cents:
Just getting CI working and sharing.


src/protocol/messages/mod.rs
```diff
@@ -33,6 +33,41 @@ pub use produce::*;
...
+        }
+    }
+}
+
+impl Default for IsolationLevel {
+    fn default() -> Self {
+        Self::ReadUncommitted
+    }
+}
+
```
crepererum, line 70: Could we have a new submodule for this (`constants.rs` maybe)? This module is already quite a grab bag for things.

tests/client.rs
```diff
@@ -23,35 +23,6 @@ async fn test_plain() {
     ClientBuilder::new(vec![connection]).build().await.unwrap();
 }
 
-#[tokio::test]
```
crepererum, line 26: I'm OK with removing this since `get_cached_leader` wasn't meant to be public anyways. On the long term it would probably nice to have two unit tests:

1. test that we can extract the partition leader from the Kafka metadata message
2. test that certain requests (produce and fetch IIRC) always use the partition leader and not some random broker

crepererum (approved):
Small nits, but I like the direction here very much :+1:

src/client/controller.rs
```diff
@@ -142,23 +120,56 @@ impl ControllerClient {
...
-    async fn get_controller_id(&self) -> Result<i32> {
-        let metadata = self.brokers.request_metadata(None, Some(vec![])).await?;
-
-        let controller_id = metadata
-            .controller_id
-            .ok_or_else(|| Error::InvalidResponse("Leader is NULL".to_owned()))?
-            .0;
+/// Takes a `request_name` and a function yielding a fallible future
+/// and handles certain classes of error
+async fn maybe_retry<B, R, F, T>(
```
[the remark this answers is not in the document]
carols10cents: Yep, the reason I didn't fully abstract `maybe_retry` is because of the errors :)

PR #80, opened 2022-01-21
refactor: Consolidate "exactly one" logic and error messages

shepmaster:


src/lib.rs
```diff
@@ -17,6 +17,21 @@
     clippy::disallowed_method
 )]
 
+trait ExactlyOne<T> {
```
crepererum, line 20: could you park this in some submodule? I would like to keep the top-level mod rather clean.

src/lib.rs
```diff
@@ -17,6 +17,21 @@
...
+    fn exactly_one(self) -> Result<T, usize>;
+}
+
+impl<T> ExactlyOne<T> for Vec<T> {
+    fn exactly_one(mut self) -> Result<T, usize> {
```
crepererum, line 25: A short docstring would be nice.

crepererum (changes requested):
Small nitpick, I like the overall code simplification here :)

crepererum (approved):

PR #82, opened 2022-01-25
feat: support gzip compression

crepererum:
While we technically don't want to emit compressed data for IOx, I'm worried that some backup / cluster tool will compress our messages anyways and we end up in some incident not being able to read the data any longer.

The primary contribution of this PR is to figure out how compression works. Adding more types can be done later and should be rather easy, since they all follow the same pattern (i.e. a compressed block that follows the number of records).


carols10cents (approved):
LGTM!

PR #83, opened 2022-01-25
fix: `retry_with_backup` takes error for "continue"

crepererum:
I think this was always meant to be this way because the request name is
passed as an argument, but this got somehow mixed up in the meantime and
now we don't probably report non-fatal errors anymore. To prevent this
in the future this tightens the type bounds a bit.


carols10cents (approved):
Great catch!

PR #86, opened 2022-01-27
fix: rare edge case in batched producer

crepererum:
I was hunting down a different issue when I found this little nugget, seems unrealistic that someone will really hit this in production but better be safe than sorry.

### DIFF
--- a/src/client/producer.rs
+++ b/src/client/producer.rs
@@ -440,11 +440,13 @@ where
                 return;
             }
         };
-        if output.is_empty() {
-            return;
-        }
 
-        let r = client.produce(output).await;
+        let r = if output.is_empty() {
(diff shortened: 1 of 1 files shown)


tustvold (approved):

src/client/producer.rs
```diff
@@ -440,11 +440,13 @@ where
                 return;
             }
         };
-        if output.is_empty() {
-            return;
-        }
 
-        let r = client.produce(output).await;
+        let r = if output.is_empty() {
```
tustvold, line 444: I think this could occur if someone flushed an empty BatchProducer??

crepererum: done

PR #87, opened 2022-01-28
chore: release 0.1.0

crepererum:
I'll tag and upload the release once this PR is merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,10 @@
 # RSKafka
 
+[![CircleCI](https://circleci.com/gh/influxdata/rskafka/tree/main.svg?style=shield&circle-token=531ba1f38035a10da6dbf7cc71e6f55eff496c70)](https://circleci.com/gh/influxdata/rskafka/tree/main)
+[![Crates.io](https://img.shields.io/crates/v/rskafka)](https://crates.io/crates/rskafka)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,5 +1,10 @@
 # RSKafka
 
+[![CircleCI](https://circleci.com/gh/influxdata/rskafka/tree/main.svg?style=shield&circle-token=531ba1f38035a10da6dbf7cc71e6f55eff496c70)](https://circleci.com/gh/influxdata/rskafka/tree/main)
```
crepererum, line 3: I case you're wondering: this is a status-only token (i.e. it can only read the status) which is required even though the project is public. It seems that w/o the token the GitHub image anonymizer (see <https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/about-anonymized-urls>) seems to run into rate limits or something resulting in a 404 when a user watches this README.

README.md
```diff
@@ -1,5 +1,10 @@
...
+[![Crates.io](https://img.shields.io/crates/v/rskafka)](https://crates.io/crates/rskafka)
+[![Documentation](https://img.shields.io/docsrs/rskafka)](https://docs.rs/crate/rskafka/latest)
```
crepererum, line 5: The docs.rs badge is currently red because the 0.0.0 placeholder that @carols10cents uploaded doesn't built. This will be fixed once we upload a real release.

tustvold (approved):

PR #91, opened 2022-01-31
feat: support LZ4 (de)compression

crepererum:


domodwyer (approved):
I love the feature flag approach 👍

PR #92, opened 2022-02-02
feat: support Snappy (de)compression

crepererum:
Snappy is a bit more messy than the other formats, but that's what IOx currently writes via rdkafka. Implementing this allows a slightly smoother transition since we don't need to drain the compressed data from our active Kafka clusters.

### DIFF
--- a/src/protocol/record.rs
+++ b/src/protocol/record.rs
@@ -24,6 +24,8 @@ use std::io::{Cursor, Read, Write};
 #[cfg(test)]
 use proptest::prelude::*;
 
+use crate::protocol::vec_builder::DEFAULT_BLOCK_SIZE;
+
 use super::{
     primitives::{Int16, Int32, Int64, Int8, Varint, Varlong},
     traits::{ReadError, ReadType, WriteError, WriteType},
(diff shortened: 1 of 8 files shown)


src/protocol/record.rs
```diff
@@ -592,6 +594,42 @@ where
 
                 records
             }
+            #[cfg(feature = "compression-snappy")]
+            RecordBatchCompression::Snappy => {
+                use snap::raw::{decompress_len, Decoder};
+
+                // Construct the input for the raw decoder.
+                let mut input = vec![];
+                reader.read_to_end(&mut input)?;
```
domodwyer, line 603: Would this allow a malicious input to trigger unbounded memory usage by feeding it `/dev/rand` for a bit?

domodwyer (approved):
I [still like](https://github.com/influxdata/rskafka/pull/91) the feature flag approach 😂

src/protocol/record.rs
crepererum: Not here because the reader is capped (it's reading from an in-memory vector that we already have at this point and this one is limited by `max_message_size` which is a protocol/client-level config). However now that I think about it we don't limit the "uncompressed size" and snappy doesn't limit that one. I'll fix that attack vector.

crepererum: fixed

PR #93, opened 2022-02-03
fix: record key/value can be NULL

crepererum:
In the NULL case (which can be produced by rdkafka) the length of these
fields is set to -1.


lukebond (approved):
i'll take your word for it :)

PR #94, opened 2022-02-03
feat: support zstd (de)compression

crepererum:
This is the last compression algorithm and since the Rust interface is really polished, this was a rather low hanging fruit.


domodwyer (approved):
That was a pleasantly small change for the functionality!

PR #96, opened 2022-02-09
fix: terminate consumer stream on `OffsetOutOfRange`

crepererum:
If we encounter `OffsetOutOfRange` this means that the start offset of
the stream lies outside of the range that Kafka knows. In this case we
should not hot-loop and try to poll new data forever, since this is very
likely NOT intended.

The only use case I could come up with is to start a stream at some "yet
to be produced" offset, but I fail to come up with a practical reasoning
behind this as well as a method to dermine the start offset for such a
use case.

So most users are likely interested to subscribe to data that already
exists within the system.


crepererum:
arg, I need to fix redpanda

tustvold (approved):

PR #97, opened 2022-02-09
feat: implement record deletion

crepererum:
We don't need record deletion in production, but having this operation available helps us to understand how Kafka behaves when data is truncated due to retention policies.

### DIFF
--- a/src/protocol/messages/delete_records.rs
+++ b/src/protocol/messages/delete_records.rs
@@ -1,0 +1,294 @@
+use std::io::{Read, Write};
+
+use crate::protocol::{
+    api_key::ApiKey,
+    api_version::{ApiVersion, ApiVersionRange},
+    error::Error,
+    messages::{
+        read_compact_versioned_array, read_versioned_array, write_compact_versioned_array,
+        write_versioned_array,
+    },
+    primitives::{CompactString, CompactStringRef, Int16, Int32, Int64, String_, TaggedFields},
+    traits::{ReadType, WriteType},
+};
+
+use super::{
+    ReadVersionedError, ReadVersionedType, RequestBody, WriteVersionedError, WriteVersionedType,
+};
+
+#[derive(Debug)]
+pub struct DeleteRequestPartition {
+    /// The partition index.
+    pub partition_index: Int32,
+
+    /// The deletion offset.
+    pub offset: Int64,
+
+    /// The tagged fields.
+    ///
+    /// Added in version 2
+    pub tagged_fields: Option<TaggedFields>,
+}
+
+impl<W> WriteVersionedType<W> for DeleteRequestPartition
+where
+    W: Write,
+{
+    fn write_versioned(
+        &self,
+        writer: &mut W,
+        version: ApiVersion,
+    ) -> Result<(), WriteVersionedError> {
+        let v = version.0 .0;
+        assert!(v <= 2);
+
+        self.partition_index.write(writer)?;
+        self.offset.write(writer)?;
+
+        if v >= 2 {
+            match self.tagged_fields.as_ref() {
+                Some(tagged_fields) => {
+                    tagged_fields.write(writer)?;
+                }
+                None => {
+                    TaggedFields::default().write(writer)?;
+                }
+            }
+        }
+
+        Ok(())
+    }
+}
+
+#[derive(Debug)]
+pub struct DeleteRequestTopic {
+    /// The topic name.
+    pub name: String_,
+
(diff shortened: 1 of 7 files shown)


tests/client.rs
```diff
@@ -338,6 +338,121 @@ async fn test_produce_consume_size_cutoff() {
...
+        .fetch_records(offset_3, 1..10_000, 1_000)
+        .await
+        .unwrap();
+    assert_eq!(
+        records,
+        vec![
+            RecordAndOffset {
+                record: record_2,
+                offset: offset_2
+            },
```
crepererum, line 443: BTW: this is technically a bug since we request data >= `offset_3`. The bug existed since forever and the next PR will fix that bug.

src/protocol/messages/delete_records.rs
```diff
@@ -0,0 +1,294 @@
...
+    primitives::{CompactString, CompactStringRef, Int16, Int32, Int64, String_, TaggedFields},
+    traits::{ReadType, WriteType},
+};
+
+use super::{
+    ReadVersionedError, ReadVersionedType, RequestBody, WriteVersionedError, WriteVersionedType,
+};
+
+#[derive(Debug)]
+pub struct DeletePartitionRequest {
```
tustvold, line 20: Perhaps something like `DeleteRecordsRequestPartition` might make it clear this is for the `DeleteRecords` API

src/protocol/messages/delete_records.rs
```diff
@@ -0,0 +1,294 @@
...
+                }
+            }
+        }
+
+        Ok(())
+    }
+}
+
+#[derive(Debug)]
+pub struct DeleteTopicRequest {
```
tustvold, line 64: Perhaps `DeleteRecordsRequestTopic` would avoid confusion with the `DeleteTopics` API

src/protocol/messages/delete_records.rs
```diff
@@ -0,0 +1,294 @@
...
+
+    /// All versions.
+    const API_VERSION_RANGE: ApiVersionRange =
+        ApiVersionRange::new(ApiVersion(Int16(0)), ApiVersion(Int16(2)));
+
+    const FIRST_TAGGED_FIELD_IN_REQUEST_VERSION: ApiVersion = ApiVersion(Int16(2));
+}
+
+#[derive(Debug)]
+pub struct DeletePartitionResponse {
```
tustvold, line 177: Same as the requests I think these should be named to make clear that they correspond to the DeleteRecords API

tests/client.rs
```diff
@@ -338,6 +338,121 @@ async fn test_produce_consume_size_cutoff() {
...
+        .await
+        .unwrap();
+    let offset_4 = offsets[0];
+
+    // delete from the middle of the 2nd batch
+    match partition_client.delete_records(offset_3, 1_000).await {
+        Ok(()) => {}
+        // Redpanda does not support deletions (https://github.com/redpanda-data/redpanda/issues/1016), but we also
+        // don't wanna skip it unconditionally
+        Err(ClientError::Request(RequestError::NoVersionMatch { .. }))
```
tustvold, line 405: :ok_hand:

tests/client.rs
```diff
@@ -338,6 +338,121 @@ async fn test_produce_consume_size_cutoff() {
...
+    let err = partition_client
+        .fetch_records(offset_2, 1..10_000, 1_000)
+        .await
+        .unwrap_err();
+    assert_matches!(
+        err,
+        ClientError::ServerError(ProtocolError::OffsetOutOfRange, _)
+    );
+
+    // fetching untouched records still works, however the middle record batch is NOT half-deleted and still contains record_2
```
tustvold, line 432: Aah yes, good-ole Kafka-style UX :laughing:

tests/client.rs
```diff
@@ -338,6 +338,121 @@ async fn test_produce_consume_size_cutoff() {
...
+    // fetching data before the record fails
+    let err = partition_client
+        .fetch_records(offset_1, 1..10_000, 1_000)
+        .await
+        .unwrap_err();
+    assert_matches!(
+        err,
+        ClientError::ServerError(ProtocolError::OffsetOutOfRange, _)
+    );
+    let err = partition_client
```
tustvold, line 423: This is wild, so if you request offset_2 it will error, but if you then fetch data it will magically return offset 2 :exploding_head:

tustvold (approved):
Looks good to me, and tests cases I wouldn't have even thought to test :laughing: 

Some minor naming nits is all

PR #100, opened 2022-02-09
feat: extend watermark API to get both low/high offset

crepererum:
Based on #99.

### DIFF
--- a/src/client/partition.rs
+++ b/src/client/partition.rs
@@ -43,6 +43,24 @@ impl Default for Compression {
     }
 }
 
+/// Which type of offset should be requested by [`PartitionClient::get_offset`].
+///
+/// # Timestamp-based Queries
+/// In theory the Kafka API would also support querying an offset based on a timestamp, but the behavior seems to be
+/// semi-defined, unintuitive (even within Apache Kafka) and inconsistent between Apache Kafka and Redpanda. So we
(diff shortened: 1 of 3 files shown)


src/client/partition.rs
```diff
@@ -585,16 +593,20 @@ fn extract_records(
     Ok(records)
 }
 
-fn build_list_offsets_request(partition: i32, topic: &str) -> ListOffsetsRequest {
+fn build_list_offsets_request(partition: i32, topic: &str, at: OffsetAt) -> ListOffsetsRequest {
+    let timestamp = match at {
+        OffsetAt::Earliest => -2,
```
tustvold, line 598: :magic_wand:

tustvold (approved):
I'm guessing this would only be useful for reporting, as the earliest offset may change before a subsequent fetch records call, leading to OffsetOutOfRange

crepererum:
> I'm guessing this would only be useful for reporting, as the earliest offset may change before a subsequent fetch records call, leading to OffsetOutOfRange

This is true for both offsets, I'll add a note/warning to the method.

crepererum:
> > I'm guessing this would only be useful for reporting, as the earliest offset may change before a subsequent fetch records call, leading to OffsetOutOfRange
> 
> This is true for both offsets, I'll add a note/warning to the method.

done

PR #104, opened 2022-02-15
feat: allow `StreamConsumer` to start at "earliest"/"latest" offset

crepererum:
This is required to provide a race-free way to start at the earliest or
latest offset, because fetching the offset and then creating a stream
might fail because in the meantime the Kafka retention policy might
already have removed the offsets in question.

### DIFF
--- a/src/client/consumer.rs
+++ b/src/client/consumer.rs
@@ -651,7 +651,7 @@ mod tests {
 
         assert_stream_pending(&mut stream).await;
 
-        // Write two records, nothing should happen as start offset is 2 (via "earlist")
+        // Write two records, nothing should happen as start offset is 2 (via "earliest")
         sender.send(record.clone()).await.unwrap();
         sender.send(record.clone()).await.unwrap();
 
@@ -694,7 +694,7 @@ mod tests {
 
         assert_stream_pending(&mut stream).await;
 
-        // Write two records, nothing should happen as start offset is 2 (via "earlist")
+        // Write two records, nothing should happen as start offset is 2 (via "latest")
         sender.send(record.clone()).await.unwrap();
         sender.send(record.clone()).await.unwrap();
 
--- a/src/client/consumer.rs
+++ b/src/client/consumer.rs
@@ -148,13 +148,13 @@ impl StreamConsumerBuilder {
     }
 }
 
-struct FetchResultInner {
+struct FetchResultOk {
     records_and_offsets: Vec<RecordAndOffset>,
     watermark: i64,
     used_offset: i64,
 }
 
-type FetchResult = Result<FetchResultInner>;
+type FetchResult = Result<FetchResultOk>;
 
 /// A trait wrapper to allow mocking
 trait FetchClient: std::fmt::Debug + Send + Sync {
@@ -252,7 +252,7 @@ impl Stream for StreamConsumer {
 
                     let (records_and_offsets, watermark) =
                         client.fetch_records(offset, bytes, max_wait_ms).await?;
-                    Ok(FetchResultInner {
+                    Ok(FetchResultOk {
                         records_and_offsets,
                         watermark,
                         used_offset: offset,
@@ -264,7 +264,7 @@ impl Stream for StreamConsumer {
 
             match (data, *this.start_offset) {
                 (Ok(inner), _) => {
-                    let FetchResultInner {
+                    let FetchResultOk {
                         mut records_and_offsets,
                         watermark,
                         used_offset,
--- a/src/client/consumer.rs
+++ b/src/client/consumer.rs
@@ -41,11 +41,12 @@ use std::ops::Range;
 use std::pin::Pin;
 use std::sync::Arc;
 use std::task::{Context, Poll};
+use std::time::Duration;
 
 use futures::future::{BoxFuture, Fuse, FusedFuture, FutureExt};
 use futures::Stream;
 use pin_project_lite::pin_project;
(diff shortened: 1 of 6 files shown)


src/client/consumer.rs
```diff
@@ -268,16 +350,20 @@ mod tests {
     struct MockFetchInner {
         batch_sizes: Vec<usize>,
         stream: mpsc::Receiver<Record>,
+        next_err: Option<Error>,
         buffer: Vec<Record>,
+        range: (i64, i64),
```
tustvold, line 355: The interaction with the offset tracking in `buffer` is perhaps a little funky. Perhaps this might be easier to understand if these were tracked automatically, and we added a delete records API... I dunno :sweat_smile:

src/client/consumer.rs
```diff
@@ -545,4 +616,104 @@ mod tests {
...
+        let mut stream = StreamConsumerBuilder::new_with_client(
+            Arc::<MockFetch>::clone(&consumer),
+            StartOffset::Earliest,
+        )
+        .with_max_wait_ms(10)
+        .build();
+
+        assert_stream_pending(&mut stream).await;
+
+        // Write two records, nothing should happen as start offset is 2 (via "earlist")
```
tustvold, line 646:
```suggestion
        // Write two records, nothing should happen as start offset is 2 (via "earliest")
```

src/client/consumer.rs
```diff
@@ -545,4 +616,104 @@ mod tests {
...
+        let mut stream = StreamConsumerBuilder::new_with_client(
+            Arc::<MockFetch>::clone(&consumer),
+            StartOffset::Latest,
+        )
+        .with_max_wait_ms(10)
+        .build();
+
+        assert_stream_pending(&mut stream).await;
+
+        // Write two records, nothing should happen as start offset is 2 (via "earlist")
```
tustvold, line 689:
```suggestion
        // Write two records, nothing should happen as start offset is 2 (via "latest")
```

tests/consumer.rs
```diff
@@ -103,3 +95,198 @@ async fn test_stream_consumer_offset_out_of_range() {
     // stream ends
     assert!(stream.next().await.is_none());
 }
+
+#[tokio::test]
+async fn test_stream_consumer_start_at_earliest() {
```
tustvold, line 100: Do we want a test that manually truncates the topic, so that the earliest offset is guaranteed to not be 0?

tustvold (approved):
Looks good to me, had some minor comments

tests/consumer.rs
crepererum: good idea, I'll add a test (that again only runs w/ Kafka since redpanda doesn't support deletes).

src/client/consumer.rs
crepererum: You mean you wanna mock a full Kafka broker? :wink:

src/client/consumer.rs
```diff
@@ -119,7 +148,13 @@ impl StreamConsumerBuilder {
     }
 }
 
-type FetchResult = Result<(Vec<RecordAndOffset>, i64)>;
+struct FetchResultInner {
```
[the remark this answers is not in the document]
crepererum: done

src/client/consumer.rs
```diff
@@ -186,38 +232,74 @@ impl Stream for StreamConsumer {
...
                         high_watermark = watermark,
-                        n_records = v.len(),
+                        n_records = records_and_offsets.len(),
                         "Received records and a high watermark",
                     );
 
+                    // Remember used offset (might be overwritten if there was any data) so we don't refetch the
+                    // earliest / latest offset for every try. Also fetching the latest offset might be racy otherwise,
+                    // since we'll never be in a position where the latest one can actually be fetched.
+                    *this.next_offset = Some(used_offset);
```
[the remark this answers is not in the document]
crepererum: So the issue here is that "latest" offset: If you have a new or empty partition, then Kafka returns the start of that empty interval (e.g. `0` if it's an empty partition) if you ask it, also see

https://github.com/influxdata/rskafka/blob/15f424f20f9da982ce79bf8fd74e4d32cfde876c/tests/client.rs#L241-L251

Now if you don't remember that offset, the time your produce something to the partition the "latest" offset (which is the latest record + 1) will also increase, so you're always chasing the tail of the partition but never fetch any records. So when you first get an answer to "what's the earliest/latest offset?" you need to remember that (except when this turns out to be wrong due to "offset out of range").

When the partition stays empty, then sure this loops forever and asks Kafka again and again for data. However this is not a hot loop since the fetch request has a timeout attached, so the broker will wait some milliseconds until it tells us "there are no records here, like the last X time you've asked me".

src/client/consumer.rs
```diff
@@ -186,38 +232,74 @@ impl Stream for StreamConsumer {
...
+                    if let Some(x) = records_and_offsets.last() {
+                        *this.next_offset = Some(x.offset + 1);
+                        this.buffer.extend(records_and_offsets.into_iter())
                     }
                     continue;
                 }
-                Err(e) => {
+                // if we don't have an offset, try again because fetching the offset is racy
+                (
+                    Err(Error::ServerError(ProtocolError::OffsetOutOfRange, _)),
```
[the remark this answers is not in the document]
crepererum: done

tests/consumer.rs
crepererum: done

PR #105, opened 2022-02-16
fix: disallow now_or_never  (and async `read`)

### LINKED ISSUE #102: Panic: just flushed
Under certain rare circumstances it seems that the `BatchProducer` panics with `just flushed`. This is this piece of code here:

https://github.com/influxdata/rskafka/blob/7c74d43f50c517c9cbdb912dc5169066330f2528/src/client/producer.rs#L439-L442

Until now I've seen that only once during benchmarking and once in prod. I'm wondering if `now_or_never()` is too aggressive.

crepererum:
Tokio might preempt tasks at any point so `now_or_never` might return
`None`. See https://tokio.rs/blog/2020-04-preemption

We could try to workaround this by using data structures from `futures`
instead of `tokio` which to not hook into the tokio cooperative
scheduling. However this has certain drawbacks

- The cooperative scheduling in tokio is actually not that bad, so we
  should rather work with it instead of against it.
- The assumption that `futures` data structures don't participate in
  cooperative scheduling is only because currently they cannot interact
  with the runtime in any way. This might change when Rust's async
  ecosystem evolves so it kinda seems like a non-future-proof
  workaround.

Fixes #102.

While I'm on it I also disallowed async `read` because it has the same issue as the `write` counterpart that we've already run into.

### DIFF
--- a/src/client/consumer.rs
+++ b/src/client/consumer.rs
@@ -450,12 +450,20 @@ mod tests {
         assert_eq!(high_watermark, 5);
 
         let (record_and_offset, high_watermark) =
-            stream.next().now_or_never().unwrap().unwrap().unwrap();
+            tokio::time::timeout(Duration::from_millis(1), stream.next())
+                .await
+                .unwrap()
+                .unwrap()
+                .unwrap();
         assert_eq!(record_and_offset.offset, 4);
         assert_eq!(high_watermark, 5);
 
         let (record_and_offset, high_watermark) =
-            stream.next().now_or_never().unwrap().unwrap().unwrap();
+            tokio::time::timeout(Duration::from_millis(1), stream.next())
+                .await
+                .unwrap()
+                .unwrap()
+                .unwrap();
         assert_eq!(record_and_offset.offset, 5);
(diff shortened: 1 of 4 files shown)


crepererum:
BTW: I haven't been able to reproduce the panic though since the task budget accounting is somewhat nested and cannot be influenced via some direct API.

src/client/consumer.rs
```diff
@@ -449,13 +449,11 @@ mod tests {
         assert_eq!(record_and_offset.offset, 3);
         assert_eq!(high_watermark, 5);
 
-        let (record_and_offset, high_watermark) =
-            stream.next().now_or_never().unwrap().unwrap().unwrap();
+        let (record_and_offset, high_watermark) = stream.next().await.unwrap().unwrap();
```
tustvold, line 452: Perhaps these could be timeout with very small timeout?

tustvold (approved):

src/client/consumer.rs
crepererum: done

PR #111, opened 2022-02-28
fix: `peek` after broadcast

crepererum:
`BroadcastOne::receive().peek()` did only return `Some(...)` if the
receiver side was actually `await`ed at least once. Thie was already
discovered and solved here:

https://github.com/influxdata/rskafka/blob/3af1939fd47f8680d40074dc3fd2e2a4a0da6b8c/src/client/producer.rs#L442-L445
https://github.com/influxdata/influxdb_iox/pull/3873/files
However in case of flushes due to "no capacity" we didn't `await`

https://github.com/influxdata/rskafka/blob/3af1939fd47f8680d40074dc3fd2e2a4a0da6b8c/src/client/producer.rs#L397

and then `peek`ed

https://github.com/influxdata/rskafka/blob/3af1939fd47f8680d40074dc3fd2e2a4a0da6b8c/src/client/producer.rs#L433

determined that we probably didn't flush the aggregator and then flushed
it again. This resulted in an empty flush and the tags that we tried to
de-aggregate didn't work (because the deaggregator that was now in the
result slot was not the one that matched our actual tag).

Instead of blindly adding more `await` calls, this PR makes the
`BroadcastOne` API a bit saver by `await`ing in `BroadcastOne::broadcast`,
so we now guarantee that `peek` always works after broadcasting.

### DIFF
--- a/src/client/producer.rs
+++ b/src/client/producer.rs
@@ -439,10 +439,7 @@ where
         // Flush data
         Self::flush_impl(&mut inner, self.client.as_ref(), self.compression).await;
 
-        // We need to poll the result slot here to make the result available via `peek`, otherwise the next thread in
-        // this critical section will flush the producer a 2nd time. We are using an ordinary `.await` call here instead
-        // of `now_or_never` because tokio might preempt us (or to be precise: might preempt polling from the result slot).
-        extract(&result_slot.await, tag)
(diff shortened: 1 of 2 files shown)


tustvold:
Superseded by #113

PR #113, opened 2022-03-02
Cancel safety

tustvold:
Builds on and is intended to replace #112. This will also replace #111 as it reworks the `BroadcastOnce` to no longer use `Shared` as its peek implementation is just unhelpful.

As reported in https://github.com/influxdata/influxdb_iox/issues/3805 we've been observing occasional panics in production. I eventually tracked this down to be due to a lack of cancellation safety within `produce`, in particular if the write request times out the future will be cancelled (dropped), causing it to drop the lock and leave the aggregator in an inconsistent state.


src/client/producer/broadcast.rs
```diff
@@ -1,76 +1,101 @@
...
-        write!(f, "ResultSlot")
-    }
+pub type Result<T, E = Error> = std::result::Result<T, E>;
+
+/// A broadcast channel that can send a single value
+///
+/// - Receivers can be created with [`BroadcastOnce::receiver`]
+/// - The value can be produced with [`BroadcastOnce::broadcast`]
+#[derive(Debug)]
+pub struct BroadcastOnce<T> {
```
tustvold, line 20: As described in #111 `Shared::peak` has some quirks when used with interior mutable futures, these can be worked around, but in the interest of keeping things simple I just switch to using Notify and a shared Option

src/client/producer.rs
```diff
@@ -374,130 +373,144 @@ where
...
+    ) -> Result<OwnedMutexGuard<ProducerInner<A>>> {
+        debug!(?client, "Flushing batch producer");
+
+        // Spawn a task to provide cancellation safety
+        let handle = tokio::spawn(async move {
+            // Swap the result slot first, this ensures that if the task panics the Drop
+            // implementation will still signal the result slot preventing flushing twice
+            let slot = std::mem::take(&mut inner.result_slot);
+
+            let (output, status_deagg) = match inner.aggregator.flush() {
```
tustvold, line 463: We still have an issue if the aggregator panics and leaves itself in an undefined state, but that's something for a future PR

src/client/producer.rs
```diff
@@ -374,130 +373,144 @@ where
...
     /// and creates a fresh result slot for future writes to use
     async fn flush_impl(
-        inner: &mut ProducerInner<A>,
-        client: &dyn ProducerClient,
+        mut inner: OwnedMutexGuard<ProducerInner<A>>,
+        client: Arc<dyn ProducerClient>,
         compression: Compression,
-    ) {
-        trace!("Flushing batch producer");
+    ) -> Result<OwnedMutexGuard<ProducerInner<A>>> {
```
tustvold, line 454: This dance passing a lock in and out of the future is a bit meh, but it is necessary for the correctness of the too large record detection - it needs to flush the aggregator, and then try to push the record again without allowing something else to push a new record in the intervening time.

src/client/producer/broadcast.rs
```diff
@@ -1,76 +1,101 @@
...
+    pub async fn receive(&self) -> Result<T> {
+        let notified = self.shared.notify.notified();
+
+        if let Some(v) = self.peek() {
+            return v;
         }
+
+        notified.await;
+
+        self.peek().unwrap()
```
crepererum, line 93:
```suggestion
        self.peek().expect("just got notified")
```

src/client/producer/broadcast.rs
```diff
@@ -81,35 +106,26 @@ mod tests {
 
         // Test simple receiver
         let broadcast: BroadcastOnce<usize> = Default::default();
-        let receiver = broadcast.receive();
+        let receiver = broadcast.receiver();
```
crepererum, line 109: I know this wasn't here before, but could we have a simple "is pending" check as well as `peek` before broadcasting, e.g. 

```suggestion
        let receiver = broadcast.receiver();
        assert!(receiver.peek().is_none());
        assert_is_pending(receiver.receive()).await;
```

src/client/producer/broadcast.rs
```diff
@@ -81,35 +106,26 @@ mod tests {
...
         broadcast.broadcast(2);
-        assert_eq!(receiver.await, 2);
+        assert_eq!(receiver.receive().await, Ok(2));
```
crepererum, line 111:
```suggestion
        assert_eq!(receiver.peek().unwrap(), Ok(2));
        assert_eq!(receiver.peek().unwrap(), Ok(2));  // can peek multiple times
        assert_eq!(receiver.receive().await, Ok(2));
        assert_eq!(receiver.receive().await, Ok(2));  // can receive multiple times
```

crepererum (approved):
Only some small nitpicks. Good job! :muscle:

src/client/producer.rs
```diff
@@ -374,130 +373,144 @@ where
     ///
     /// # Cancellation
     ///
-    /// The returned future is not cancellation safe, if it is dropped the record
-    /// may or may not be published
+    /// The returned future is cancellation safe in that it won't leave the [`BatchProducer`]
+    /// in an inconsistent state, however, the provided data may or may not be produced
```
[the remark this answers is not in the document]
tustvold: Correct, cancellation should not be able to side-effect on others anymore

PR #115, opened 2022-03-07
fix: make `Messenger::request` cancellation-safe

### LINKED ISSUE #103: Frequent connection losses
On the producer side, we see loads (~once per minute) connection losses in prod, like:

`level=info msg="request encountered non-fatal error - backing off" e="Request error: Connection is poisened: Cannot read framed message: Cannot read data: early eof" request_name=produce backoff_secs=0 target="rskafka::backoff"`

They are automatically retried, but it would be nice if we could prevent this from happening THAT often.

crepererum:
This avoids that we write partial messages to the underlying TCP/TLS
connection and mess up message framing. This will nearly always lead to
broker-side errors and the broker closes the connection.

Fixes #103.

### DIFF
--- a/src/messenger.rs
+++ b/src/messenger.rs
@@ -941,15 +941,24 @@ mod tests {
 
         let messenger = Messenger::new(tx_front, 1_000);
 
+        // create two barriers:
+        // - pause: will be passed after 3 bytes were sent by the client
+        // - continue: will be passed to continue client->broker traffic
+        //
+        // The barriers do NOT affect broker->client traffic.
+        //
+        // The sizes of the barriers is 2: one for the network simulation task and one for the main/control thread.
         let network_pause = Arc::new(Barrier::new(2));
         let network_pause_captured = Arc::clone(&network_pause);
+        let network_continue = Arc::new(Barrier::new(2));
+        let network_continue_captured = Arc::clone(&network_continue);
         let handle_network = tokio::spawn(async move {
             // Need to split both directions into read and write halfs so we can run full-duplex in two loops. Otherwise
             // the test might deadlock even though the code is just fine (TCP is full-duplex).
             let (mut rx_middle_read, mut rx_middle_write) = tokio::io::split(rx_middle);
             let (mut tx_middle_read, mut tx_middle_write) = tokio::io::split(tx_middle);
 
-            let direction1 = async {
+            let direction_client_broker = async {
                 for i in 0.. {
                     let mut buf = [0; 1];
                     rx_middle_read.read_exact(&mut buf).await.unwrap();
@@ -957,11 +966,12 @@ mod tests {
 
                     if i == 3 {
                         network_pause_captured.wait().await;
+                        network_continue_captured.wait().await;
                     }
                 }
             };
 
-            let direction2 = async {
+            let direction_broker_client = async {
                 loop {
                     let mut buf = [0; 1];
                     tx_middle_read.read_exact(&mut buf).await.unwrap();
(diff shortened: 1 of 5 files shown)


src/messenger.rs
```diff
@@ -922,6 +932,122 @@ mod tests {
         assert_eq!(actual, resp);
     }
 
+    #[tokio::test]
+    async fn test_cancel_request() {
```
tustvold, line 936: I think this test would be easier to follow if the loops were broken up into separate functions, but it's fine as is also

src/messenger.rs
```diff
@@ -922,6 +932,122 @@ mod tests {
...
+        // pause it after 3 bytes are sent by the client.
+        let (tx_front, rx_middle) = tokio::io::duplex(1);
+        let (tx_middle, mut rx_back) = tokio::io::duplex(1);
+
+        let messenger = Messenger::new(tx_front, 1_000);
+
+        let network_pause = Arc::new(Barrier::new(2));
+        let network_pause_captured = Arc::clone(&network_pause);
+        let handle_network = tokio::spawn(async move {
+            // Need to split both directions into read and write halfs so we can run full-deplux in two loops. Otherwise
```
tustvold, line 947:
```suggestion
            // Need to split both directions into read and write halfs so we can run full-duplex in two loops. Otherwise
```

src/messenger.rs
```diff
@@ -922,6 +932,122 @@ mod tests {
...
+    #[tokio::test]
+    async fn test_cancel_request() {
+        // Use a "virtual" network between a simulated broker and a client. The network is intercepted in the middle to
+        // pause it after 3 bytes are sent by the client.
+        let (tx_front, rx_middle) = tokio::io::duplex(1);
+        let (tx_middle, mut rx_back) = tokio::io::duplex(1);
+
+        let messenger = Messenger::new(tx_front, 1_000);
+
+        let network_pause = Arc::new(Barrier::new(2));
```
tustvold, line 944: Perhaps an explanation of why 2?

src/messenger.rs
```diff
@@ -922,6 +932,122 @@ mod tests {
...
+                .unwrap();
+        })
+        .fuse();
+
+        {
+            // cancel when we exit this block
+            pin_mut!(task_to_cancel);
+
+            // write exactly 3 bytes via the client and then cancel the task.
+            futures::select_biased! {
```
tustvold, line 1025: I think this is racy, as network_pause.wait() could be reached allowing the network to continue and the task to finish, before this task is scheduled and exits the select block. I'm not sure that select_biased will prevent this, as even if network_pause triggers the wakeup, task_to_cancel will still be polled first

src/messenger.rs
```diff
@@ -922,6 +932,122 @@ mod tests {
...
+            tokio::select! {
+                _ = direction1 => {}
+                _ = direction2 => {}
+            }
+        });
+
+        // simulated broker, just reads messages and answers w/ "api versions" responses
+        let handle_broker = tokio::spawn(async move {
+            for correlation_id in 0.. {
+                rx_back.read_message(1_000).await.unwrap();
```
tustvold, line 981: Could possibly check that the message is an ApiVersions request

tustvold (approved):
Looks good to me, I think the test has a race, but otherwise :muscle:

src/messenger.rs
crepererum: done

src/messenger.rs
crepererum: done

PR #116, opened 2022-03-08
chore: Add rustfmt.toml

pierwill:
This is helpful to ensure IDEs (Emacs in my case) use the right edition when calling out to `rustfmt`.


crepererum (approved):
I always thought that these tools also use `Cargo.toml`, but I guess there is no harm in adding this file.

crepererum:
I'll close and re-open this PR since I've changed some CircleCI settings to build your forked branch.

Issue #101, opened 2022-02-14
Convert `as` conversions to `From`/`TryFrom`


crepererum:
The `as` conversions in our protocol code our quite dangerous and I think we might have a few integer overflow bugs in there. In general I think we should replace all `as` conversions with `From` and `TryFrom` to make it clear if the conversion can fail or not. There is also a clippy lint [`as_conversions`](https://rust-lang.github.io/rust-clippy/master/index.html#as_conversions) which should be added to our code base to prevent regressions.

pierwill:
👋 I'd like to work on this.

PR #117, opened 2022-03-09
fix: Use a `try_from` instead of a cast

pierwill:
This is part of https://github.com/influxdata/rskafka/issues/101.


src/protocol/frame.rs
```diff
@@ -70,7 +69,7 @@ where
             });
         }
 
-        let mut buf = vec![0u8; len as usize];
+        let mut buf = vec![0u8; len];
```
pierwill, line 72: `len` is already a `usize` here.

crepererum (approved):

PR #118, opened 2022-03-10
fix: Remove `as` conversions in protocol/primitives.rs

pierwill:
Part of #101.

### DIFF
--- a/src/protocol/primitives.rs
+++ b/src/protocol/primitives.rs
@@ -236,9 +236,9 @@ where
         let mut shift = 0;
         loop {
             reader.read_exact(&mut buf)?;
-            let c = buf[0];
+            let c: u64 = buf[0].into();
 
-            res |= ((c as u64) & 0x7f) << shift;
+            res |= (c & 0x7f) << shift;
             shift += 7;
 
             if (c & 0x80) == 0 {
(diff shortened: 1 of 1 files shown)


src/protocol/primitives.rs
```diff
@@ -552,11 +556,13 @@ where
     W: Write,
 {
     fn write(&self, writer: &mut W) -> Result<(), WriteError> {
-        UnsignedVarint(self.0.len() as u64).write(writer)?;
+        let len = u64::try_from(self.0.len()).map_err(WriteError::Overflow)?;
+        UnsignedVarint(len).write(writer)?;
```
crepererum, line 560: Just a small note because we had a chat about integer sizes yesterday:

`.len()` is usually `usize` which on 64bit systems is the same as `u64`. So this is a no-op on 64bit systems (basically all desktops, laptops, servers and about all smartphones nowadays). For 32bit systems (embedded) this also wouldn't be an issue because we would convert to a wider integer. It would only be an issue if `usize` is 128bit (the only system I know that would allow that is the RISC-V 128bit architecture and there are no chips or compilers available for it) AND when we have more than `2^64 - 1` elements in that list (which would also be insane). So I think this is somewhat of an impossible error case, but on the other hand it also doesn't hurt since the method in question already returns a `Result`.

crepererum (approved):

PR #120, opened 2022-03-14
fix: Remove `as` conversions in protocol/record.rs

pierwill:
Part of work on #101.

### DIFF
--- a/src/protocol/record.rs
+++ b/src/protocol/record.rs
@@ -104,8 +104,8 @@ where
     fn read(reader: &mut R) -> Result<Self, ReadError> {
         // length
         let len = Varint::read(reader)?;
-        let len = usize::try_from(len.0).map_err(|e| ReadError::Malformed(Box::new(e)))?;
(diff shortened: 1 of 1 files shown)


src/protocol/record.rs
```diff
@@ -105,7 +105,7 @@ where
         // length
         let len = Varint::read(reader)?;
         let len = usize::try_from(len.0).map_err(|e| ReadError::Malformed(Box::new(e)))?;
-        let reader = &mut reader.take(len as u64);
+        let reader = &mut reader.take(u64::try_from(len).map_err(ReadError::Overflow)?);
```
crepererum, line 108: I think you can get away w/ a single conversion here:

```rust
let len = Varint::read(reader)?;
let len = u64::try_from(len.0).map_err(|e| ReadError::Malformed(Box::new(e)))?;
let reader = &mut reader.take(len);
```

crepererum (approved):

PR #121, opened 2022-03-14
fix: Remove more `as` conversions in `protocol::messages`

pierwill:
Part of #101.

### DIFF
--- a/src/protocol/messages/mod.rs
+++ b/src/protocol/messages/mod.rs
@@ -187,8 +187,7 @@ fn read_compact_versioned_array<R: Read, T: ReadVersionedType<R>>(
     match len {
         0 => Ok(None),
         n => {
-            let len = usize::try_from(n - 1)
-                .map_err(ReadError::Overflow)?;
+            let len = usize::try_from(n - 1).map_err(ReadError::Overflow)?;
             let mut builder = VecBuilder::new(len);
             for _ in 0..len {
                 builder.push(T::read_versioned(reader, version)?);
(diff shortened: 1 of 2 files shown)


src/protocol/messages/mod.rs
```diff
@@ -137,7 +137,7 @@ fn read_versioned_array<R: Read, T: ReadVersionedType<R>>(
             format!("Invalid negative length for array: {}", l).into(),
         ))),
         _ => {
-            let len = len as usize;
+            let len = usize::try_from(len).map_err(ReadError::Overflow)?;
```
pierwill, line 140: Should this be `.map_err(|e| ReadVersionedError::ReadError(ReadError::Overflow(e)))?`?

crepererum: I think `thiserror` takes care of that itself:

https://github.com/influxdata/rskafka/blob/2c13f4bea7cb8ec3d0f3a5ab4eee98cf393ca50d/src/protocol/messages/mod.rs#L40-L44

See the `#[from]` macro.

src/protocol/messages/mod.rs
```diff
@@ -187,7 +187,8 @@ fn read_compact_versioned_array<R: Read, T: ReadVersionedType<R>>(
     match len {
         0 => Ok(None),
         n => {
-            let len = (n - 1) as usize;
+            let len = usize::try_from(n - 1)
+                .map_err(|e| ReadVersionedError::ReadError(ReadError::Overflow(e)))?;
```
crepererum, line 191:
```suggestion
            let len = usize::try_from(n - 1)
                .map_err(ReadError::Overflow)?;
```

Should be sufficient, see discussion above.

crepererum (approved):

PR #123, opened 2022-03-29
fix: TLS feature flag

domodwyer:
Some parts of the code were referencing types behind feature flags, and CI wasn't catching it - this PR corrects both.

---

* fix: missing TLS feature gates (1c0eb13)

      Adds missing cfg(tls) annotations to code blocks that require it, allowing
      cargo check --all-targets to pass.

* ci: check all targets with no features (04f6863)

      Adds --all-targets to the "no default features" build, which causes it to
      validate the flags used in #[cfg(test)] gated blocks too.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -620,7 +620,7 @@ mod tests {
 
         async fn connect(
             &self,
-            _tls_config: Option<Arc<rustls::ClientConfig>>,
+            _tls_config: TlsConfig,
             _socks5_proxy: Option<String>,
             _max_message_size: usize,
         ) -> Result<Arc<Self::R>> {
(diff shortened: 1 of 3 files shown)


src/connection.rs
```diff
@@ -620,7 +620,7 @@ mod tests {
 
         async fn connect(
             &self,
-            _tls_config: Option<Arc<rustls::ClientConfig>>,
+            _tls_config: TlsConfig,
```
crepererum, line 623: I was briefly wondering why this code compiled before, but it turns out that this is test code as well and therefore wasn't checked via `cargo build --no-default-features`. Good catch!

PR #124, opened 2022-03-30
feat: split produce batches for MESSAGE_TOO_LARGE errors

domodwyer:
This PR allows a batched `produce()` call to try and automatically handle `MESSAGE_TOO_LARGE` errors by splitting the aggregated batch in two, and submitting each half individually. 

While most incidents of this error are due to bad configuration, it can occur even with correctly tuned message sizes due to the aggregator using the approximated record sizes. A `WARN` log event is emitted each time this automatic splitting is needed to surface any potential misconfiguration to the operator.

This automatic splitting has pros, and cons:

* Pro: client doesn't have to care about batch sizing approximation going wrong
* Pro: the produce caller retains the original `Record`, rather than giving up ownership - this makes resubmitting / error handling possible for the caller.
* Con: extra pointer indirection & ref counting (negligent in the grand scheme of things)
* Con: partial writes can be infrequently triggered in the `MESSAGE_TOO_LARGE` path, as are spurious produce errors

---

* refactor: wrap record key/value in Arc (43b776e)

      Allows "copies" of the key/value bytes to be held for produce request retries
      without having to actually copy the underlying buffer, at the minor cost of
      extra pointer indirection.

* refactor: lend Records to produce() calls (39bac2d)

      Changes the ProducerClient::produce() call to take a slice of Records rather
      than a Vec, allowing the caller to retain the original buffer and easily
      provide a subset slice of the original buffer as needed.

* refactor: let MockClient return response sequences (9b7ecf1)

      This change affects test code only, changing the MockClient to return a 
      sequence of configured errors, rather than the same error all the time.

      As an example, this can be used to construct a test that observes an error,
      followed by a different error, followed by a success - the previous mock was
      restricted to always returning the same response.

* feat: split produce batch for MESSAGE_TOO_LARGE (99e15b3)

      Aggregating the records for produce requests is approximate, and as such 
      sometimes may result in a batch larger than the configured maximum - if the
      size of the batch exceeds the broker's maximum allowed message size
      (message.max.bytes) a MESSAGE_TOO_LARGE error is returned.

      At this point and prior to this commit, the whole batch was dropped and the
      caller has to retry, hoping the retries do not batch into a too-large batch
      once again. After this commit, exactly one attempt is made to
      automatically/internally split the batch in half and attempt to concurrently
      produce each half transparently to the caller. A message is logged at WARN to
      help the user identify and tune this config to avoid this.

      The two downsides of this are:

         * Unpredictable latency if a batch needs splitting and re-submitting
         * A partial write may occur (writing one half succeeds, other fails)

      These downsides are offset by the fact the MESSAGE_TOO_LARGE should be a rare
      event, making the latency hit equally rare. Succeeding in producing one batch
      but not the other is even rarer still, but does result in an error being
      returned to the client even though have the records were successfully
      produced. The caller had no control over which writes were in which aggregated
      batches anyway, so has no expectation of two writes succeeding/failing
      together (atomicity).

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ let partition_client = client
 // produce some data
 let record = Record {
     key: None,
-    value: Some(b"hello kafka".to_vec()),
+    value: Some(Arc::new(b"hello kafka".to_vec())),
     headers: BTreeMap::from([
         ("foo".to_owned(), b"bar".to_vec()),
     ]),
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ use rskafka::{
     record::Record,
 };
 use time::OffsetDateTime;
-use std::collections::BTreeMap;
+use std::{collections::BTreeMap, sync::Arc};
 
 // setup client
 let connection = "localhost:9093".to_owned();
@@ -71,7 +71,7 @@ let record = Record {
     ]),
     timestamp: OffsetDateTime::now_utc(),
 };
-partition_client.produce(vec![record], Compression::default()).await.unwrap();
+partition_client.produce(&[record], Compression::default()).await.unwrap();
 
 // consume data
 let (records, high_watermark) = partition_client
--- a/benches/write_throughput.rs
+++ b/benches/write_throughput.rs
@@ -82,7 +82,7 @@ pub fn criterion_benchmark(c: &mut Criterion) {
 
                     for _ in 0..iters {
                         client
-                            .produce(vec![record.clone()], Compression::NoCompression)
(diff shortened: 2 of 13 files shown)


crepererum:
For the record: @domodwyer and I agreed that this approach doesn't really work if the aggregator just produces as single record (which is very likely for the IOx use case). Instead we would need to have some inverse operation to the aggregator (=splitter). Implementing this would be nice, but it needs some API design so the whole producer framework doesn't explode in terms of complexity. So we will postpone that for later.

PR #129, opened 2022-04-21
WIP: Add Sasl authentication - PLAIN

### LINKED ISSUE #125: Support for authorization methods.
Hi there,

I am attempting to use `rskafka` to connect to a Kafka cluster using authorization using this [tutorial](https://medium.com/swlh/a-free-apache-kafka-cloud-service-and-how-to-quickly-get-started-with-it-8f14520fff35) (for NodeJS).
Is authz supported at all? If so how do I provide such data to `ClientBuilder`? I cannot find any example of this in the code. 

thanks in advance!

ramrengaswamy:
Closes #125 

Describe your proposed changes here.
- Added support for `PLAIN` SASL auth.
- Added a couple of new requests `SaslHandshakeRequest` and `SaslAuthenticateRequest`
- Added a couple of primitives `Bytes` (not used) and `CompactBytes`.
- Added a `SaslConfig` struct and that can be set in the `ClientBuilder` and used during connection establishment.
- [X] I've read the contributing section of the project [CONTRIBUTING.md](https://github.com/influxdata/rskafka/blob/main/CONTRIBUTING.md).
- [X] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed).


src/protocol/messages/sasl_auth.rs
```diff
@@ -0,0 +1,206 @@
...
+    }
+}
+
+impl<R> ReadVersionedType<R> for SaslHandshakeRequest
+where
+    R: Read,
+{
+    fn read_versioned(reader: &mut R, version: ApiVersion) -> Result<Self, ReadVersionedError> {
+        let v = version.0 .0;
+        assert!(v <= 1);
```
crepererum, line 39: This is not in line w/ `API_VERSION_RANGE`: you don't support version 0.

src/protocol/messages/sasl_auth.rs
```diff
@@ -0,0 +1,206 @@
...
+where
+    W: Write,
+{
+    fn write_versioned(
+        &self,
+        writer: &mut W,
+        version: ApiVersion,
+    ) -> Result<(), WriteVersionedError> {
+        let v = version.0 .0;
+        assert!(v == 2);
```
crepererum, line 147: Why is version 1 not supported?

Edit: reading <https://kafka.apache.org/protocol#sasl_handshake> again I think we can also support version 0 for the `SaslAuthenticate` message.

src/protocol/primitives.rs
```diff
@@ -478,6 +478,61 @@ where
     }
 }
 
+#[derive(Debug)]
+pub struct Bytes(pub Vec<u8>);
```
crepererum, line 482: Looks good. Please copy over the docs from <https://kafka.apache.org/protocol#protocol_types> and add a roundtrip test.

src/messenger.rs
```diff
@@ -469,6 +470,14 @@ where
...
         Err(SyncVersionsError::NoWorkingVersion)
     }
+
+    pub async fn sasl_handshake(&self, sasl_config: &SaslConfig) -> Result<(),  RequestError> {
+        let req = SaslHandshakeRequest::new();
+        self.request(req).await?;
+        let req = SaslAuthenticateRequest::new(sasl_config);
+        self.request(req).await?;
+        Ok(())
+    }
```
crepererum, line 480: So if there is a `SaslConfig`, then we need to call this each time after `sync_versions`. I think you can just grep throw the code and see where is this done. Maybe you wanna make the two methods private and instead offer a `setup` method that does both, so we don't forget to call one or the other a certain callsides.

crepererum:
@ramrengaswamy this is a good start :+1:

src/protocol/messages/sasl_auth.rs
crepererum: The sentence you're quoting is about the `SaslHandshakeRequest` (for which I agree that we should only support version 1), not about the `SaslAuthenticateRequest`. The versions for these two request-response types are different. The version selection will be done by the `Messenger`.

ramrengaswamy: I realize that your comment was regarding `SaslAuthenticate`. The issue I ran in to here was that the type of the fields got modified through the different versions. For e.g. `auth_bytes` was `BYTES` in version 0 but become `COMPACT_BYTES` from version 1 onwards. Is there an example for this is handled anywhere else in the library?

src/connection.rs
```diff
@@ -105,6 +108,11 @@ impl ConnectionHandler for BrokerRepresentation {
 
         let messenger = Arc::new(Messenger::new(BufStream::new(transport), max_message_size));
         messenger.sync_versions().await?;
+        if let Some(sasl_config) = sasl_config {
```
ramrengaswamy, line 111: @crepererum I didn't refactor this into a separate `setup` function because I wasn't sure `SaslConfig` should leak into the `Messenger`. Right now `SaslConfig` is known to `transport, connection and ClientBuilder`.

src/protocol/messages/sasl_auth.rs
crepererum: I see. TBH we don't have a beautiful solution of it. I would go for `Bytes` as a type within the struct and then do some type conversion during read/write. See this example:

https://github.com/influxdata/rskafka/blob/ad8de3ccd7183f0d380ef2bd65a6fc1ea74fffb9/src/protocol/messages/delete_records.rs#L63-L66

Here the `name` is an ordinary string, but is compact after a certain version. This is then handled here:

https://github.com/influxdata/rskafka/blob/ad8de3ccd7183f0d380ef2bd65a6fc1ea74fffb9/src/protocol/messages/delete_records.rs#L89-L93

I think you might want to introduce `CompactBytesRef` for that, so you don't have to copy the payload during type conversion. This is similar how it is done for `CompactString`:

https://github.com/influxdata/rskafka/blob/ad8de3ccd7183f0d380ef2bd65a6fc1ea74fffb9/src/protocol/primitives.rs#L362-L365

https://github.com/influxdata/rskafka/blob/ad8de3ccd7183f0d380ef2bd65a6fc1ea74fffb9/src/protocol/primitives.rs#L392-L399

(note how the "write" operation is only implemented once, because `CompactString` can use `CompactStringRef`)

https://github.com/influxdata/rskafka/blob/ad8de3ccd7183f0d380ef2bd65a6fc1ea74fffb9/src/protocol/primitives.rs#L401-L417

src/connection.rs
crepererum: Looks good to me. I've just checked: this is the only place where we call `sync_versions` and also the place where we set up SOCKS5 and TLS. So I think this is a good spot to call `sasl_handshake` :+1:

src/protocol/messages/sasl_auth.rs
ramrengaswamy: Neat! I will follow a similar pattern.

ramrengaswamy:
@crepererum ... Ready for another look.
This is my first real rust program so also let me know if some code doesn't look `"rusty"` :)
I will fix those.

src/client/mod.rs
```diff
@@ -77,12 +79,22 @@ impl ClientBuilder {
...
     }
 
+    /// Setup SASL username and password. Mechanism is assumed to be PLAIN.
+    pub fn sasl_config(mut self, username: &str, password: &str) -> Self {
+        self.sasl_config = Some(SaslConfig::Plain{
+            username: username.to_string(),
+            password: password.to_string(),
+        });
+        self
+    }
```
crepererum, line 89: I would take `SaslConfig` as an argument here so we can extend the config in the future w/o overloading the `ClientBuilder` w/ too many methods.

src/connection/transport/sasl.rs
```diff
@@ -0,0 +1,25 @@
...
+
+impl SaslConfig {
+    pub fn auth_bytes(&self) -> Vec<u8> {
+        match self {
+            Self::Plain { username, password } => {
+                let mut auth: Vec<u8> = Vec::new();
+                auth.push(0);
+                auth.extend(username.bytes());
+                auth.push(0);
+                auth.extend(password.bytes());
```
crepererum, line 14: Can you add a link/reference for this binary format?

I think for `PLAIN` it's OK to hand-roll this, in the future however I would like to offload this to another crate because I don't wanna deal w/ all the SASL crypto stuff in rskafka.

src/connection/transport/sasl.rs
```diff
@@ -0,0 +1,25 @@
+#[derive(Debug, Clone)]
+pub enum SaslConfig {
+    Plain { username: String, password: String },
+}
+
+impl SaslConfig {
+    pub fn auth_bytes(&self) -> Vec<u8> {
```
crepererum, line 7:
```suggestion
    pub(crate) fn auth_bytes(&self) -> Vec<u8> {
```

This is not a public API that falls under semantic versioning and that users should rely on. Same goes for `fn mechanism`.

crepererum (commented):
Looks solid :+1: 

Now how do we test this? :grin:

src/connection/transport/sasl.rs
ramrengaswamy: Are you okay depending on a pure rust crate such as this one? https://docs.rs/sasl/latest/sasl/
The other SASL packages are rust wrappers around C implementations.

src/protocol/messages/sasl_msg.rs
```diff
@@ -0,0 +1,260 @@
...
+    const API_VERSION_RANGE: ApiVersionRange =
+        ApiVersionRange::new(ApiVersion(Int16(1)), ApiVersion(Int16(1)));
+    const FIRST_TAGGED_FIELD_IN_REQUEST_VERSION: ApiVersion = ApiVersion(Int16(3));
+}
+
+#[derive(Debug)]
+pub struct SaslHandshakeResponse {
+    /// The error code, or 0 if there was no error.
+    ///
+    /// Added in version 0.
```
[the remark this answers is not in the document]
ramrengaswamy: Got rid of it to keep it consistent with the rest of the codebase.

src/client/mod.rs
ramrengaswamy: Had to `pub use crate::connection::SaslConfig` in `client/mod.rs` to publicly expose `SaslConfig`

ramrengaswamy:
I am currently testing these changes against a simple application that I am running against a Confluent stream. 
I am not sure if it is possible to run a local Kafka instance with SASL auth enabled. Will look into it.

ramrengaswamy:
@crepererum - Seems that the local Kafka and Redpanda have sasl enabled. So I added a `test_sasl` that will trigger the SASL handshake and authenticate exchanges in the integration tests.

src/messenger.rs
```diff
@@ -469,6 +469,18 @@ where
...
+        mechanism: &str,
+        auth_bytes: Vec<u8>,
+    ) -> Result<(), RequestError> {
+        let req = SaslHandshakeRequest::new(mechanism);
+        self.request(req).await?;
+        let req = SaslAuthenticateRequest::new(auth_bytes);
+        self.request(req).await?;
+        Ok(())
+    }
 }
```
crepererum, line 484: I was really surprised that this "just works" in our CI tests. In short: It doesn't. You have to check the error codes within the messages (`SaslHandshakeResponse::error_code` and `SaslAuthenticateResponse::error_code`) and return an error if the codes are set.

You may wonder how I found that out that quickly: Wireshark:

image

ramrengaswamy: Ugh, can't believe I missed that! Fixing.

ramrengaswamy:
@crepererum tl;dr - I got the sasl integration test working locally but ran into an issue getting to work on CircleCI.

To get Sasl integration test to work I made the following changes:

1. Added a new `SECURE` listener to all the brokers and configured its security to `SASL_PLAINTEXT`.
2. Added a `kafka_jaas.conf` file to get the brokers to boot up successfully and handle simple password auth.
3. Introduced a new environment variable `KAFKA_SASL_CONNECT` that is used by `test_sasl` to connect to the `SECURE` listener.

I have not used CircleCI before. It seems there is no way to volume mount into a Docker container.
So we would have to use this [workaround](https://circleci.com/docs/2.0/building-docker-images/#mounting-folders) to copy `kafka_jaas.conf` into CircleCI and make it accessible to the brokers.

crepererum:
@ramrengaswamy Didn't have time to look into this today, sorry. I'll be out of office for two weeks, I'll get back to this on the 16th of May.

seijik42:
I tried to contribute this by fixing the test for Redpanda but seems like they don't support SASL/PLAIN at the moment. (only SASL/SCRAM)
https://docs.redpanda.com/docs/manage/security/authentication/#saslscram
Is to ok to simply skip `test_sasl` when envvar `TEST_BROKER_IMPL=redpanda` ?

toondaey:
Is there any progress on this?

crepererum:
@toondaey I think the latest attempt is #198 . If you could pick that one up (just create a new PR) and make sure that the tests also run in CircleCI, I think we could easily drive it to completion.

toondaey:
> @toondaey I think the latest attempt is #198 . If you could pick that one up (just create a new PR) and make sure that the tests also run in CircleCI, I think we could easily drive it to completion.

Sure. I'll give it a whirl.

Neustradamus:
Linked to:
- https://github.com/scram-sasl/info/issues/1
- @ramrengaswamy: https://github.com/influxdata/rskafka/pull/129
- @seijik42: https://github.com/influxdata/rskafka/pull/198
- @toondaey: https://github.com/influxdata/rskafka/pull/216

PR #135, opened 2022-05-12
test: SOCKS5 in CI and docker compose files

### LINKED ISSUE #110: Enable SOCKS5 integration test
Add some SOCKS5 proxy to our CI and enable

https://github.com/influxdata/rskafka/blob/3af1939fd47f8680d40074dc3fd2e2a4a0da6b8c/tests/client.rs#L118-L122

This should be gated by some environment variable like `SOCKS5_ENDPOINT` so that developers don't need to have a SOCKS5 proxy running locally if they don't work on that particular feature.

crepererum:
Closes #110.


mario-s (approved):
Nice! Then I will just close my PR, once this one is merged.

PR #138, opened 2022-05-13
ci: run fuzzer regularly

crepererum:


domodwyer (approved):
Nice careful decoding dance!

Well worth the effort of setting up fuzzing 👍

PR #141, opened 2022-05-30
fix: fix potential OOM involving snappy data, found by fuzzer

crepererum:
Fuzzer on `main` is currently failing, see <https://app.circleci.com/pipelines/github/influxdata/rskafka/1252/workflows/99a91b16-68d1-4f74-975c-ad7d5cae53c5/jobs/5657/steps>.


alamb (approved):
Does it need a test too? Or maybe fuzzing is sufficient

domodwyer (approved):
No more spam!

crepererum:
> Does it need a test too?

I think I can extract a test case :+1:

crepererum:
added a test

PR #146, opened 2022-07-13
test: ensure that record timestamps can be produced/consumed out of order

crepererum:


src/client/partition.rs
```diff
@@ -399,7 +399,6 @@ fn build_produce_request(
 
     // TODO: Retry on failure
 
-    // TODO: Verify this is the first timestamp in the batch and not the min
```
tustvold, line 402: I'm guessing the conclusion is it is the first and not the min?

tustvold (approved):

src/client/partition.rs
crepererum: yes, ensured by the cross-tests with the other implementations.

Issue #147, opened 2022-07-14
`PartitionClient` may silently read from replica


crepererum:
When a cluster re-balances, the existing partition leader may become a replica, but will not actively report that to the `PartitionClient`. Even worse, the read lag may trigger this Redpanda quirk even under Kafka:

https://github.com/influxdata/rskafka/blob/822186d69dcc41f0e02785228a273504e069baaf/src/client/partition.rs#L149-L159

PR #150, opened 2022-07-14
fix: remove potentially buggy Redpanda quirk

crepererum:
Newer versions of Redpanda do not require the `OffsetOutOfBounds`
quirk anymore, so let's remove it. See #147 on why this is a good idea.


crepererum:
Looking at the socks5 test failure...

tustvold:
Fix in https://github.com/redpanda-data/redpanda/pull/4201 included in v22.1.1

PR #160, opened 2022-07-27
feat: extend `ServerError` with helpful data

### LINKED ISSUE #148: More helpful `OffsetOutOfRange` error
`OffsetOutOfRange` may happen on two _sides_:

1. The offset that we are trying to read was deleted (either via the API or via the retention policy)
2. The offset does not exist yet (or is gone for another reason, e.g. because a topic was re-created).

Both cases are reported using the very same error code, but the FETCH response should have enough information (log start, high watermark) to determine the difference.

crepererum:
~On top of #159.~

Closes #148.


src/client/error.rs
```diff
@@ -15,8 +40,30 @@ pub enum Error {
...
+        error_message: Option<String>,
+
+        /// Additional context that we can tell the user about.
+        context: Option<ServerErrorContext>,
+
+        /// Additional payload that we can provide the user.
+        payload: Option<ServerErrorPayload>,
+
+        /// Flags if the error was generated by the client to simulate some server behavior or workaround a bug.
+        is_virtual: bool,
```
tustvold, line 65: What is the intended use-case for this?

src/client/error.rs
```diff
@@ -15,8 +40,30 @@ pub enum Error {
...
+    )]
+    ServerError {
+        /// Protocol-level error message.
+        protocol_error: ProtocolError,
+
+        /// Server message provided by the broker, if any.
+        error_message: Option<String>,
+
+        /// Additional context that we can tell the user about.
+        context: Option<ServerErrorContext>,
```
tustvold, line 59:
```suggestion
        request: Option<ServerErrorContext>,
```

src/client/error.rs
```diff
@@ -3,6 +3,31 @@ use thiserror::Error;
 pub use crate::messenger::RequestError;
 pub use crate::protocol::error::Error as ProtocolError;
 
+/// Context for [`Error::ServerError`].
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum ServerErrorContext {
```
tustvold, line 9:
```suggestion
pub enum RequestContext {
```

src/client/error.rs
```diff
@@ -3,6 +3,31 @@ use thiserror::Error;
...
+    Partition(String, i32),
+}
+
+/// Payload for [`Error::ServerError`].
+///
+/// This is data that the server sent and that is still usable despite the error.
+#[derive(Debug)]
+#[allow(missing_copy_implementations)] // wanna extend this later
+#[non_exhaustive]
+pub enum ServerErrorPayload {
```
tustvold, line 20:
```suggestion
pub enum ServerErrorResponse {
```

src/client/error.rs
```diff
@@ -15,8 +40,30 @@ pub enum Error {
...
+        protocol_error: ProtocolError,
+
+        /// Server message provided by the broker, if any.
+        error_message: Option<String>,
+
+        /// Additional context that we can tell the user about.
+        context: Option<ServerErrorContext>,
+
+        /// Additional payload that we can provide the user.
+        payload: Option<ServerErrorPayload>,
```
tustvold, line 62:
```suggestion
        response: Option<ServerErrorResponse>,
```

src/client/error.rs
```diff
@@ -15,8 +40,30 @@ pub enum Error {
...
+        context,
+        payload,
+        is_virtual
+    )]
+    ServerError {
+        /// Protocol-level error message.
+        protocol_error: ProtocolError,
+
+        /// Server message provided by the broker, if any.
+        error_message: Option<String>,
```
tustvold, line 56: I wonder if this could be merged into `ServerErrorResponse`?

tustvold (commented):
It wasn't initially clear why the fields were separated, I've left some clarifications that I think might help?

src/client/error.rs
crepererum: debugging mostly

src/client/error.rs
crepererum: I think most newer Kafka messages now have an error code and a human-readable string (i.e. this message). The response is mostly what else we could extract. So yeah, we could merge it but I guess the resulting enum will rather unhelpful because in many cases we don't have anything else than the error message.

src/client/error.rs
crepererum: done

src/client/error.rs
crepererum: done

src/client/error.rs
crepererum: done

src/client/error.rs
crepererum: done

crepererum:
@tustvold PTAL

PR #164, opened 2022-08-10
feat: metadata cache for leader discovery

domodwyer:
This PR halves the number of requests needed by the `PartitionClient` for leadership discovery in the common/happy path.

Ideally we would perform a metadata request, and then use that to initialise all the `PartitionClient` a user needs in one go and hand them to the user - however because a `PartitionClient` is lazily initialised by a call to [`Client::partition_client()`] at any point in time, it is not possible to perform batch initialisation in this way (but is a slightly nicer UX).

Instead the approach this PR takes is to maintain a metadata cache (invalidated & updated automatically as a side effect of normal usage) and use the cached metadata to optimistically select a partition's leader. This is then validated by contacting the potential leader to assert it indeed is the actual leader (existing code). I expect this cache will also help reduce recovery times when a cluster disruption occurs.

Testing shows this can lead to ~15% speed-up when attempting to initialise a large number of `PartitionClient` instances, increasing as the I/O latency increases. Test performed on a local, single node, no-replica cluster with no active workloads, so the "real world" improvement is likely much greater.

[`Client::partition_client()`]:https://github.com/influxdata/rskafka/blob/706cc67c6cbed84651a4647f21caf497fea52fc6/src/client/mod.rs#L118-L129

---

* feat: metadata cache for leader discovery (706cc67)

      Adds a MetadataCache, used by the BrokerCache to optimistically halve the
      number of broker requests made during leader discovery.

### DIFF
--- a/src/client/controller.rs
+++ b/src/client/controller.rs
@@ -7,7 +7,9 @@ use tracing::{debug, error, info};
 use crate::{
     backoff::{Backoff, BackoffConfig},
     client::{Error, Result},
-    connection::{BrokerCache, BrokerConnection, BrokerConnector, MessengerTransport},
+    connection::{
+        BrokerCache, BrokerConnection, BrokerConnector, MessengerTransport, MetadataLookupMode,
+    },
     messenger::RequestError,
     protocol::{
         error::Error as ProtocolError,
@@ -93,7 +95,7 @@ impl ControllerClient {
         // Request an uncached, fresh copy of the metadata.
         let metadata = self
             .brokers
-            .request_metadata(None, Some(vec![]), false)
+            .request_metadata(MetadataLookupMode::ArbitraryBroker, Some(vec![]))
             .await?;
 
         let controller_id = metadata
--- a/src/client/controller.rs
+++ b/src/client/controller.rs
@@ -80,12 +80,21 @@ impl ControllerClient {
                 }),
             }
         })
-        .await
+        .await?;
+
+        // Refresh the cache now there is definitely a new topic to observe.
+        let _ = self.brokers.refresh_metadata().await;
+
+        Ok(())
     }
 
     /// Retrieve the broker ID of the controller
     async fn get_controller_id(&self) -> Result<i32> {
-        let metadata = self.brokers.request_metadata(None, Some(vec![])).await?;
+        // Request an uncached, fresh copy of the metadata.
+        let metadata = self
+            .brokers
+            .request_metadata(None, Some(vec![]), false)
+            .await?;
 
         let controller_id = metadata
             .controller_id
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -12,6 +12,7 @@ use crate::{
 pub mod consumer;
 pub mod controller;
 pub mod error;
+pub(crate) mod metadata_cache;
 pub mod partition;
 pub mod producer;
 
@@ -129,7 +130,15 @@ impl Client {
 
     /// Returns a list of topics in the cluster
     pub async fn list_topics(&self) -> Result<Vec<Topic>> {
-        let response = self.brokers.request_metadata(None, None).await?;
+        // Do not used a cached metadata response to satisfy this request, in
+        // order to prevent:
+        //
+        //  * Client creates a topic
+        //  * Client calls list_topics() and does not see new topic
+        //
+        // Because this is an unconstrained metadata request (all topics) it
+        // will update the cached metadata entry.
+        let response = self.brokers.request_metadata(None, None, false).await?;
 
         Ok(response
             .topics
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -4,7 +4,7 @@ use thiserror::Error;
 
 use crate::{
     client::partition::PartitionClient,
-    connection::{BrokerConnector, TlsConfig},
+    connection::{BrokerConnector, MetadataLookupMode, TlsConfig},
     protocol::primitives::Boolean,
     topic::Topic,
 };
@@ -138,7 +138,10 @@ impl Client {
         //
         // Because this is an unconstrained metadata request (all topics) it
         // will update the cached metadata entry.
-        let response = self.brokers.request_metadata(None, None, false).await?;
+        let response = self
+            .brokers
+            .request_metadata(MetadataLookupMode::ArbitraryBroker, None)
+            .await?;
 
         Ok(response
             .topics
--- a/src/client/partition.rs
+++ b/src/client/partition.rs
@@ -1,7 +1,9 @@
 use crate::{
     backoff::{Backoff, BackoffConfig},
     client::error::{Error, RequestContext, Result},
-    connection::{BrokerCache, BrokerConnection, BrokerConnector, MessengerTransport},
+    connection::{
+        BrokerCache, BrokerConnection, BrokerConnector, MessengerTransport, MetadataLookupMode,
+    },
     messenger::RequestError,
     protocol::{
         error::Error as ProtocolError,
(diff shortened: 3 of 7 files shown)


src/client/partition.rs
```diff
@@ -295,24 +299,47 @@ impl BrokerCache for &PartitionClient {
...
+                Err(Error::InvalidResponse(format!(
+                    "Partition leader {} not found in metadata response",
+                    leader
+                )))
+            }
+            Err(e) => {
+                self.brokers.invalidate_metadata_cache();
+                Err(e.into())
+            }
+        }?;
```
domodwyer, line 325: Any failure to connect to the cached "leader" invalidates the cache.

src/client/mod.rs
```diff
@@ -129,7 +130,15 @@ impl Client {
 
     /// Returns a list of topics in the cluster
     pub async fn list_topics(&self) -> Result<Vec<Topic>> {
```
domodwyer, line 132: `list_topics()` will always refresh the cached entry as a "free" side effect, as will creating a topic, or an explicit call to `refresh_metadata()`

src/client/partition.rs
```diff
@@ -295,24 +299,47 @@ impl BrokerCache for &PartitionClient {
...
         //
         // This does not seem to be required for redpanda.
-        let leader_self = self.get_leader(Some(Arc::clone(&broker))).await?;
+        //
+        // Because this check requires the most up-to-date metadata from a specific broker, do not accept a cached
+        // metadata response.
+        let leader_self = self.get_leader(Some(Arc::clone(&broker)), false).await?;
         if leader != leader_self {
+            // The cached metadata identified an incorrect leader - it is stale and should be refreshed.
+            self.brokers.invalidate_metadata_cache();
```
domodwyer, line 341: If the cached leader was not the leader, the cache is invalidated.

src/connection.rs
```diff
@@ -172,13 +181,47 @@ impl BrokerConnector {
...
+        // NOTE: no serialisation / locking is imposed over the cache during
+        // this call  - multiple parallel calls to this function will still
+        // perform multiple requests until the cache is populated. However, the
+        // Client initialises this cache at construction time, so unless
+        // invalidated, there will always be a cached entry available.
+        if use_cache {
+            assert!(
+                broker_override.is_none(),
+                "cannot ask for cached metadata from specific broker"
+            );
```
domodwyer, line 219: The cache maintains metadata acquired from an arbitrary broker, so it is an error to request a cached entry from a specific broker.

src/connection.rs
```diff
@@ -192,12 +235,22 @@ impl BrokerConnector {
         let response =
             metadata_request_with_retry(broker_override, &request, backoff, self).await?;
 
+        // If the request was for a full, unfiltered set of topics, cache the
+        // response for later calls to make use of.
+        if request.topics.is_none() {
+            self.cached_metadata.update(response.clone());
+        }
+
```
domodwyer, line 243: The cache is updated any time a suitable metadata request is issued.

domodwyer (commented):
It might also be worth implementing a batch `ProducerClient` initialisation method on the `Client` :+1:

src/client/partition.rs
```diff
@@ -295,24 +299,47 @@ impl BrokerCache for &PartitionClient {
...
+        // that happen after we connect to a leader (as advised by another broker) which doesn't know about its assigned
+        // partition yet. The metadata query below seems to result in an LeaderNotAvailable in this case and is retried
+        // by the layer above.
         //
         // This does not seem to be required for redpanda.
-        let leader_self = self.get_leader(Some(Arc::clone(&broker))).await?;
+        //
+        // Because this check requires the most up-to-date metadata from a specific broker, do not accept a cached
+        // metadata response.
+        let leader_self = self.get_leader(Some(Arc::clone(&broker)), false).await?;
```
domodwyer, line 338: This check is the remaining request in the discovery phase and is somewhat unavoidable as-is.

I wonder if we can optimistically assume the leader is correct, trap on the `UnknownTopicOrPartition` error described here, and use it to trigger a re-run of discovery (with fresh metadata) to eliminate this pessimistic check?

crepererum:
Will review tomorrow.

src/client/partition.rs
crepererum: I think there is a similar question for #147. There we read from a broker that knows the partition (hence there is no `UnknownTopicOrPartition`) but is only a read replicate and hasn't caught up with the partition leader yet (i.e. we get an `OffsetOutOfRange`).

I agree that we should probably fix both by trapping these errors and assume that the user didn't want to run into them deliberately, invalidate the cache and retry. However we don't wanna retry forever because they are valid actual errors. So this needs a bit more sophisticated retry logic and I suggest we do that in a later PR.

src/connection.rs
crepererum: Maybe instead of using two bi-states (`Option` and `bool`) with one invalid combination use an enum to encode the tri-state ("use cache", "fetch fresh from arbitrary broker", "fetch fresh from specific broker"). The same goes for `get_leader`.

src/client/controller.rs
```diff
@@ -80,12 +80,21 @@ impl ControllerClient {
                 }),
             }
         })
-        .await
+        .await?;
+
+        // Refresh the cache now there is definitely a new topic to observe.
+        let _ = self.brokers.refresh_metadata().await;
```
crepererum, line 86: That assumes two things:

- the creator of the topic is actually interested in producing / consuming them, which is NOT true for pure admin clients
- we immediately have usable metadata available. I think for Kafka+Zookeeper the partition leader election will take some time, so it is very likely that the response we cache here doesn't contain any usable information at all.

Overall this seems like a premature optimization. I would rather leave it out and expose some "refresh metadata" call on top-level `Client` that a user CAN call after creating loads of topics and before creating all the clients in hope that this speeds up things (although I'm not really sure that this actually helps).

crepererum (changes requested):
Good catch, although 15% speed up for client creation (which is rather a one-time thingy for a usually long-running process) seems a bit negligible on its own. Hope you find more optimizations that stack on top :wink: :brain: 

I have two nitpicks (the tri-state and the premature optimization) that I would like to see fixed (or defended) before merging.

src/client/controller.rs
domodwyer: I included it beacuse:

* It is a common pattern to create a topic and immediately initialise clients for it ([we do it!](https://github.com/influxdata/influxdb_iox/blob/b21799acaea38be920b1cc5942a3243abd37dbbf/write_buffer/src/kafka/mod.rs#L64-L78))
* The single extra request is preferable over initialising 100 `PartitionClient` with no cached metadata (if all created quickly they race to populate the cache)
* Producing/consuming clients > admin clients
* I'd hope we can hide the implementation details of caching from the user and not have them know they need to call `refresh()` to avoid a perf footgun

Ultimately I added this because I was creating a topic and initialising 100 `PartitionClient` for it in a local test and saw no performance improvements at all - turns out, it was because the cache wasn't populated with the topic I had just created, so caching was useless! I figured this was relatively common, so stuck this refresh here - I'm not married to it though.

domodwyer:
> 15% speed up for client creation (which is rather a one-time thingy for a usually long-running process) seems a bit negligible on its own

It's 15% on localhost with effectively no latency / completely unloaded brokers - I expect it'll be "worth more" in real life usage, but I wanted a number to check I hadn't optimised it worse :D 

Also I expect this will help in another, less obvious way: recovery thundering hurd. When a broker goes offline or the cluster is disturbed in some way and leader discovery needs to re-run, I suspect not hammering it with 2N requests might help recovery time overall - I'd love to get it down to 1 request and remove the validation!

src/client/partition.rs
domodwyer: Yeah definitely not wanting to do it here...

I think ultimately performing an optimistic leader discovery would be much better for multiple reasons, but the subtleties sound tricky. It sounds like simply bounding the number of retries to something sensible might be an easy solution. I'll create a ticket and see how much pain is left in prod to see if it justifies working on it :+1:

crepererum (approved):
Thank you!

domodwyer:
Reporting back now I have real-world numbers!

Prior to this PR, on a busy prod workload, we have observed producers taking ~2m45s to complete discovery across 100 partitions (with significant contention) against a loaded kafka cluster. After this PR, the latest measurement was at 1s :tada:

PR #165, opened 2022-08-11
feat: support pre-warming PartitionClient

domodwyer:
This exposes a method that can be used to drive leader discovery _before_ the first call to `produce()` on a `PartitionClient`. This allows us to "pre-warm" the cached broker connection to the leader and avoid doing it when we send through the first write.

Ideally we would do this automatically when initialising the `PartitionClient`, but that would be a breaking change  as [the constructor](https://github.com/influxdata/rskafka/blob/b612f08cefe92c26d54db61044d219d994e671e6/src/client/mod.rs#L119-L129) would need to become async - happy to change it if a breaking change is acceptable :+1:

I wasn't really sure how to test this :shrug:

---

* feat: support pre-warming PartitionClient (b612f08)

      Adds a refresh_leader() method to force the PartitionClient through the 
      discovery process.

      This will ensure a cached connection to the leader is (likely) already 
      established by the time the first produce() call is issued, preventing the
      first produce() calls (multiple of which may be queued up on the discovery
      mutex) from experiencing excessively high latency.

* refactor: pre-warm PartitionClient connection (387a96d)

      Change the Client::partition_client() constructor to always perform leader
      detection and establish a connection to the broker immediately, instead of
      differing the work until the first produce() call.

      This eliminates the unusually high latency of the first produce() call.

      BREAKING CHANGE: PartitionClient constructor is now async

* refactor: retry failed leader discovery (e060f60)

      Adds a retry loop (w/ backoff) over the leader discover when initialising a
      PartitionClient.

### DIFF
--- a/src/client/partition.rs
+++ b/src/client/partition.rs
@@ -115,6 +115,17 @@ impl PartitionClient {
         self.partition
     }
 
+    /// Re-run the leader discovery process for this client, and establish a
+    /// connection to the leader.
+    pub async fn refresh_leader(&self) -> Result<()> {
+        // Remove the current broker connection, if any.
+        *self.current_broker.lock().await = None;
+        // And acquire a new one, forcing discovery and establishing a cached
+        // connection to the leader.
+        self.get().await?;
+        Ok(())
+    }
+
     /// Produce a batch of records to the partition
     pub async fn produce(
         &self,
--- a/src/client/partition.rs
+++ b/src/client/partition.rs
@@ -117,7 +117,7 @@ impl PartitionClient {
 
     /// Re-run the leader discovery process for this client, and establish a
     /// connection to the leader.
-    pub async fn refresh_leader(&self) -> Result<()> {
+    pub(crate) async fn refresh_leader(&self) -> Result<()> {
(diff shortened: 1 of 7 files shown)


domodwyer:
I have made the breaking change proposed in the description - now all users will have a fully-primed `PartitionClient` without any extra hops 👍

src/client/partition.rs
```diff
@@ -115,6 +115,17 @@ impl PartitionClient {
...
+    /// Re-run the leader discovery process for this client, and establish a
+    /// connection to the leader.
+    pub(crate) async fn refresh_leader(&self) -> Result<()> {
+        // Remove the current broker connection, if any.
+        *self.current_broker.lock().await = None;
+        // And acquire a new one, forcing discovery and establishing a cached
+        // connection to the leader.
+        self.get().await?;
+        Ok(())
+    }
```
crepererum, line 127: Since this is only used in combination w/ `PartitionClient::new` I would inline this function into `new`. This also spares you the `*self.current_broker.lock().await = None;` bit.

crepererum (approved):

PR #172, opened 2022-08-24
test: improve testing config

crepererum:
Instead of having X different macros and control flow exists, let's have
one mechanism to read env variables and decide if a test shall run or
not.

This also moves the "which broker implementation supports which features"
from a bunch of environment variables into a nice enum. This will
greatly help when we expand the set of admit APIs, see #157 for example.


domodwyer (approved):
Makes sense to me - I didn't immediately understand why we had to treat the impls differently, but I see redpanda lacks delete support 🤷

PR #173, opened 2022-08-25
perf: async flushing & buffering + lock contention

domodwyer:
This is a big one that I couldn't break up into anything smaller - sorry!

Happy to talk through the changes if needed, if not, grab a cuppa! 🍵 Large portions of the change are comments, actual LoC change is a small percentage.

---

The main aims of this PR are outlined below, in order of priority:

### Make aggregating batches, and writing already full batches asynchronous w.r.t each other

Currently this is not the case. A call to [`produce()`] with space in the aggregator goes like this:

1. [Acquires the lock covering the aggregator state](https://github.com/influxdata/rskafka/blob/26af7701f6d1d64982f779fe56f3b470cf5c1751/src/client/producer.rs#L406)
2. [Successfully adds the data to the aggregator](https://github.com/influxdata/rskafka/blob/26af7701f6d1d64982f779fe56f3b470cf5c1751/src/client/producer.rs#L413)
3. [Wait for the linger timeout](https://github.com/influxdata/rskafka/blob/26af7701f6d1d64982f779fe56f3b470cf5c1751/src/client/producer.rs#L436)
4. [Acquire the lock over the aggregator state again (same lock as 1.)](https://github.com/influxdata/rskafka/blob/26af7701f6d1d64982f779fe56f3b470cf5c1751/src/client/producer.rs#L441)
5. **Whilst holding the lock**, [flush the batch](https://github.com/influxdata/rskafka/blob/26af7701f6d1d64982f779fe56f3b470cf5c1751/src/client/producer.rs#L456).

Because the lock is held while the flush happens, all callers to [`produce()`] begin queuing up, even though their payloads would be **buffered**. This means they needlessly incur the latency of the protocol overhead/serialisation/network I/O, only to then have to try and stuff as many payloads as possible into the buffer before the next 5ms linger timer blocks aggregation again.

**In an environment where writing to kafka incurs a RTT of ~50ms, we spend ~5ms aggregating payloads before blocking for ~50ms, repeated indefinitely.**

While most Kafka writes are quick, we observe long tail latencies:

image

After the changes, flushing happens asynchronously, without having to hold a lock over the aggregator state. This allows ~5ms of payloads to be batched up and dispatched, and immediately the next batch begins being built up, allowing near continuous buffer building. This should reduce the overall latency of calls to `produce()` and increase the throughput of messages to Kafka.

I've been careful to construct the buffering / flushing such that there is no need for an async mutex - given that the mutex is in the buffering hot path, replacing it with a non-async `parking_lot` mutex (which has far lower overhead) reduces the latency of calls that simply add the payload to an im-memory buffer. This should slightly increase the number of messages we can buffer in the ~5ms linger period.

### Flush result readers do not contend with buffering `produce()` callers

Previously all calls to `produce()` would lead the task to waiting for either the linger to fire, or the flush to complete:

https://github.com/influxdata/rskafka/blob/26af7701f6d1d64982f779fe56f3b470cf5c1751/src/client/producer.rs#L435-L438

However given the low default linger time (5ms) relative to the average RTT latency of a flush (especially under high load), it's probable most, if not all waiters will observe their linger timer expiring before the flush completes and the result is broadcast to the `result_slot`. I've observed many `Already flushed` logs in our prod, with no `Insufficient capacity in aggregator - flushing` to cause the flush in the first place, indicating this happening quite often.

Once the linger timer fires, all the callers proceed to queue up to [acquire the aggregator mutex](https://github.com/influxdata/rskafka/blob/26af7701f6d1d64982f779fe56f3b470cf5c1751/src/client/producer.rs#L441) again, unnecessarily serialising readers when their result is available and adding further contention with callers of `produce()` hoping to buffer their payloads in parallel.

After this PR, only one caller will attempt to acquire the aggregator mutex to initiate a flush, and the rest of the callers can wait on their result broadcast only.

Removing this lock contention allows for more `produce()` calls to buffer their payloads within the 5ms linger, increasing the message sizes over the wire.

Reading the broadcast result has also been made cheaper by reducing contention of the `BroadcastOnce` primitive with many readers, switching an exclusive mutex to a many-readers lock - they now no longer have to queue up to sequentially read (+ clone) the results, which should marginally reduce the per-call latency to `produce()` too.

### Useless Wakeups

In the above linger-expired-acquire-lock situation, the readers all .await on the timer, then .await on the aggregator mutex, only to then queue up on the mutex synchronising the [peek()](https://github.com/influxdata/rskafka/blob/26af7701f6d1d64982f779fe56f3b470cf5c1751/src/client/producer.rs#L448) call. This causes many useless wakeups only to immediately hit the next `Pending` await point, as well as unnecessary thread blocking queuing on the non-async results mutex.

Reducing the number of awaits in this path should help reduce the (small) tokio overhead incurred when parking/waking tasks.

[`produce()`]: https://github.com/influxdata/rskafka/blob/26af7701f6d1d64982f779fe56f3b470cf5c1751/src/client/producer.rs#L400

## Behavioural Changes

None of these changes necessitate a breaking external API change :tada:

#### Back-pressure

Previously rskafka would allow one single buffer flush to be in-flight. This (very!) effectively applied back-pressure to clients, slowing them down.

After this change, there **is no back-pressure** and limiting the number of outstanding `produce()` calls is the responsibility of the caller. Apps can throw as much as they want at rskafka, and it'll do the best it can to keep up, with `produce()` latencies increasing as load increases.

It is, however, relatively simple to add back-pressure to rskafka using a semaphore if desired.

#### Tokio

This requires the [`is_finished()`](https://docs.rs/tokio/1.19.2/tokio/task/struct.JoinHandle.html#method.is_finished) method on the task `JoinHandle`, which was [added in `1.19`](https://github.com/tokio-rs/tokio/releases/tag/tokio-1.19.0) and required a min tokio version bump.

---

* perf: async batch flushing & lock contention (9e9d848)

      Rewrites the management of the aggregation buffer to enable asynchronous 
      flushing of batched writes, while continuing to build up the next buffer of
      writes in parallel.

      A behavioural change of note: previously there was (synchronous) back-pressure
      applied when a batch was being flushed, and there could only ever be at most 1
      batch of writes outstanding. After this commit this back-pressure is removed
      as a side effect of async flushing, and there is no upper bound on the number
      of in-flight batch writes enforced. Back-pressure can easily be placed back
      into rskafka if it is deemed the responsibility of rskafka, rather than the
      caller, to limit the number of outstanding produce() calls.

      This commit also eliminates lock contention between the callers attempting to
      add to the batch, and those that are waiting for their write result, and also
      minimises the spurious task wakeups caused by all callers waiting on the
      linger timer, before (likely) moving on to wait on the async mutex, before
      finally being able to read their write result over the broadcast channel (also
      a mutex). Now one caller performs a linger dance, while all the others wait
      exclusively on their broadcast handle for the results, avoiding having to wait
      on the aggregation mutex.

* perf: use RwLock for BroadcastOnce (1d5be60)

      Prior to this commit, all the readers (holders of BroadcastOnceReceiver) would
      contend with each other to read the value that has been broadcast. The
      BroadcastOnce type restricts usage to one writer, with many readers, making
      the case of >1 reader contending to read the result common by design.

      A RwLock eliminates the reader contention, preventing the readers from queuing
      up to read their value - it is now a pair of cheap atomic inc/sub to read the
      value, in parallel, for each reader.


src/client/batch.rs
```diff
@@ -0,0 +1,178 @@
+use std::sync::Arc;
```
crepererum, line 1: This file should live in `rskafka::client::producer` since it's producer-specific.

src/client/producer.rs
```diff
@@ -323,64 +326,246 @@ impl ProducerClient for PartitionClient {
...
+            if token != self.flush_clock {
+                debug!(client=?self.client, "spurious batch flush call");
+                return Ok(());
+            }
+        }
+
+        debug!(client=?self.client, "flushing batch");
+
+        // Remove the batch, temporarily swapping it for a None until a new
+        // batch is built.
```
crepererum, line 478: This is NOT panic-safe. If `batch.background_flush` which calls `aggregator.flush` panics than the entire aggregator is poisoned forever[^1]. I think we should allow the aggregator to panic and only error the affected inputs instead of all future actions.

[^1]: This BTW is the reason why the stdlib mutexes implement poisoning.

src/client/producer.rs
```diff
@@ -323,64 +326,246 @@ impl ProducerClient for PartitionClient {
...
+    ///
+    /// If the caller provides a `flusher_token`, the batch flush is conditional
+    /// on the token matching. If the token does not match, the batch the caller
+    /// is attempting to flush has already been flushed, and this call is a NOP.
+    fn flush(&mut self, flusher_token: Option<usize>) -> Result<()> {
+        // If this caller is is intending to conditionally flush a specific
+        // batch, verify this BatchBuilder is the batch it is indenting to
+        // flush.
+        if let Some(token) = flusher_token {
+            if token != self.flush_clock {
```
crepererum, line 469: Can this only happen when there are manual flushes involved or are there other cases as well?

src/client/batch.rs
```diff
@@ -0,0 +1,178 @@
...
+            // Broadcast an empty result set to satisfy the aggregation
+            // contract.
+            self.results.broadcast(Ok(Arc::new(AggregatedStatus {
+                aggregated_status: vec![],
+                status_deagg,
+            })));
+            return FlushResult::Ok(Self::new(self.aggregator), None);
+        }
+
+        let handle = tokio::spawn({
```
crepererum, line 158: Why do we need yet another task here? Isn't there already one that drives the linger and the flushing? Or is this so we have also cancellation safety if there are manual flushes?

src/client/producer.rs
```diff
@@ -209,18 +209,22 @@ use std::time::Duration;
```
crepererum, line 209: There is a flow-chart on the top of this file (GitHub doesn't let me comment there). This one only shows the user-visible data flow and this is all good. From an rskafka developer PoV, I would welcome if the interplay between the producer, the aggregator and the batches (which now contain their own background task) would be illustrated somewhere.

src/client/producer.rs
```diff
@@ -323,64 +326,246 @@ impl ProducerClient for PartitionClient {
...
-
     client: Arc<dyn ProducerClient>,
 
-    inner: Arc<Mutex<ProducerInner<A>>>,
+    /// A list of (potentially) outstanding flush tasks.
+    ///
+    /// These may or may not yet be complete, and completed flush tasks are
+    /// removed from this list when adding new flush tasks or manually flushing
+    /// with a call to [`BatchProducer::flush()`].
+    pending_flushes: Vec<JoinHandle<()>>,
```
crepererum, line 369: We should abort all background tasks if all user callsides are gone or cancelled:

```rust
impl Drop for ProducerInner {
    fn drop(&mut self) {
        for handle in self.pending_flushes.drain(..) {
            handle.abort();
        }
    }
}
```

Technically I'm wondering if we should even cancel a specific flush tasks if all callsides for that specific task are gone :thinking: Because what's the point of sending this data then.

BTW: You may assume that the `PartitionClient` is cancellation-safe (everything else would be a bug). The underlying `Messenger` ensure that we never send partial messages to the broker.

crepererum (commented):
I like the split between "aggregate" and "send batch" that you perform here. I have a couple of questions and remarks.

src/client/batch.rs
domodwyer: 786d6e1

src/client/producer.rs
domodwyer: That's right - I did consider panics here, but took the view that a panicking aggregator is an exceptional event (i.e. not an recoverable error) and should be propagated up. 

We're effectively implementing poisoning here by having the next call `take()` a `None` and exploding ourselves - I think it's reasonable to expect panics to be exceptional / fatal, but open to ideas.

src/client/producer.rs
domodwyer: This block of code intentionally doesn't care why the tokens don't match!

But to answer properly, this can happen two ways:

* A manual call to [`BatchProducer::flush()`] causes the buffer to be flushed early. The caller doing the linger either:
    * [Observes the result of the flush early, before the linger expires][early], or
    * The linger fires after the manual `flush()`, but before the result is broadcast. In this case, [this conditional flush][after] becomes a NOP.

* At least two writes to an aggregator are made; the first caller starts lingering, and the second caller [observes a full aggregator][full] and flushes the batch before the linger expires. The subsequent flush attempt by the linger is a NOP (which may also take the same early flush path as above).

[full]:https://github.com/influxdata/rskafka/blob/82862df069978a6fcca52567ddf16bda85d4a8cc/src/client/producer.rs#L430-L457
[after]:https://github.com/influxdata/rskafka/blob/82862df069978a6fcca52567ddf16bda85d4a8cc/src/client/producer.rs#L654
[early]:https://github.com/influxdata/rskafka/blob/82862df069978a6fcca52567ddf16bda85d4a8cc/src/client/producer.rs#L663
[`BatchProducer::flush()`]: https://github.com/influxdata/rskafka/blob/82862df069978a6fcca52567ddf16bda85d4a8cc/src/client/producer.rs#L681

src/client/producer.rs
domodwyer: c0ba668

src/client/producer.rs
domodwyer: I'll knock one up this afternoon and PR it - I think it'll take me a bit of digging around (you're right, the interplay is complex!)

Issue #144, opened 2022-06-09
i dont want import time::OffsetDateTime


joshua86z:
```
let record = Record {
            key: None,
            value: Some(b"hello kafka".to_vec()),
            headers: BTreeMap::from([]),
            timestamp: OffsetDateTime::now_utc(),
        };
```
i am already import `chrono`
i dont want import `time`
😃

crepererum:
1. We re-export `time` for you, so you don't need to add it to your direct dependencies. Just `use rskafka::time;`.
2. You can convert between `time` and `chrono` types manually (e.g. by using unix timestamps)
3. Supporting multiple time libraries is out of range for this crate. Should the overwhelming majority of rust community swing back from `time` to `chrono` (after doing the opposite half a year ago while `chrono` was basically abandoned), we may reconsider our choice.

PR #176, opened 2022-09-06
refactor: replace `time` w/ `chono`

crepererum:
Looking at IOx, we currently have 3 time-like dependencies:

- time 0.1: this is a backwards-compat dependencies of chrono 0.4 and
  will be removed in chrono 0.5
- time 0.3: only used by rskafka
- chrono 0.4: used by everything else

Let's remove time 0.3 from that list. Also it seems that chrono gains
more traction recently, so IMHO it's the safer bet.

Also see #144.


domodwyer (approved):
Nice!

PR #178, opened 2022-10-17
ci: update fuzz job scheduling

crepererum:
CircleCI phases out the currently used mechanism, so we need to migrate it to use project-wide triggers.


domodwyer (approved):
Looks good minus lint failure.

lukebond (approved):
docs reference for the API change for those reading along: https://circleci.com/docs/scheduled-pipelines/#migrate-scheduled-workflows

PR #180, opened 2022-10-26
feat: allow setting a custom client ID

crepererum:


mkmik (approved):
ideally we should just get config via the standard kafka map<string,string>; client.id would be one of the many kafka config options conveyed with that method

crepererum:
> ideally we should just get config via the standard kafka map<string,string>; client.id would be one of the many kafka config options conveyed with that method

image

You mean the "replicate whatever the default Java client in some version X does"? :wink:

Issue #182, opened 2022-10-27
Support throttling


crepererum:
Properly react to `throttle_time_ms` in protocol responses. Currently we ignore that field, so if a `fetch` request is throttled this leads to some bizarre errors:

`invalid response: Expected a single topic in response, got 0`

PR #183, opened 2022-10-27
refactor: remove interior mutable from `Messenger::version_ranges`

crepererum:
Small clean-up before I start to work on #182.


domodwyer (approved):
Accidental `mut`?

lukebond (approved):

crepererum:
> Accidental `mut`?

Semi. It is mutable while bootstrapping the connection, but during bootstrap we also have a mutable reference to the `Messenger`. The versions should not change/re-sync while the connection is in use.

Issue #62, opened 2022-01-14
Cache Invalidation Race


tustvold:
Currently in various places we cache connections, and invalidate them on error. This is currently racy, and may end up invalidating a connection different from the one that error-ed, if something else concurrently invalidated the cache and then fetched a new connection. 

This is unlikely to be a problem in practice, but the fix is simple - pass the `BrokerConnection` to the invalidate call and only invalidate if `Arc::ptr_eq` is true

PR #186, opened 2022-10-28
fix: (potential) metadata cache invalidation race

crepererum:
Under certain conditions it MAY happen that metadata cache invalidation races, e.g.:

| Thread 1     | Thread 2     |
| ------------ | ------------ |
| get metadata |              |
|              | get metadata |
| invalidate   |              |
| get metadata |              |
|              | invalidate   |

Here the 2nd invalidation is unnecessary. It's not a correctness issue, but we throw away cached data that is perfectly fine.

To work around this, we now track a "generation" for data that we get from the metadata cache and check this during invalidation requests.

Ref #62. (not closing, need to do the same for the broker connection cache)


domodwyer (approved):
👌

PR #230, opened 2024-01-23
feat: Minor improvements to error visibility

einarmo:
This PR makes a few small changes:

 - Changes the derived display impl of `BackoffError` to also include the `source` error.
 - Removes a few unhelpful TODO's, which are no longer correct.
 - Exposes `connection::Error` as `ConnectionError`. Currently it is not visible at all, since `connection` is a private module.
 - Captures errors from broker connection and passes that along too.

Nothing should be breaking or particularly intrusive.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -1,5 +1,6 @@
 use async_trait::async_trait;
 use rand::prelude::*;
+use std::fmt::Display;
 use std::ops::ControlFlow;
 use std::sync::Arc;
 use thiserror::Error;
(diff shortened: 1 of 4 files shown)


src/connection.rs
```diff
@@ -485,16 +486,24 @@ where
...
                 return ControlFlow::Break(connection);
             }
 
-            let err = Box::<dyn std::error::Error + Send + Sync>::from(
-                "Failed to connect to any broker, backing off".to_string(),
-            );
+            // errors should always contain at least 1 element here.
+            let errors_string = errors
+                .into_iter()
+                .map(|e| e.to_string())
```
crepererum, line 500: I suggest you create a helper type here to delay the error formatting until it is actually needed:

```rust
struct MultiError(Vec<Box<dyn std::error::Error + Send + Sync>);

impl std::error::Error for MultiError {...}
```

einarmo: Sure, that's probably better. Maybe it would even be more useful with


```
struct MultiError<T: std::error::Error + Send + Sync + 'static>(Vec<T>)
```

einarmo: Though since we're sticking it in a `Box<dyn ...>` anyway I guess it doesn't make much of a difference in this case.

crepererum:
tests need a fix and the commit message shall follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) (just prefix it with `refactor: `)

einarmo:
I have three separate commits, as they are technically three separate changes, but I'll squash them if you prefer.

crepererum (approved):

crepererum:
Thank you and sorry for the review delay.

PR #245, opened 2024-08-05
chore: rust 1.80

crepererum:
- update MSRV
- drop `once_lock`
- drop `async_trait`

### DIFF
--- a/tests/test_helpers.rs
+++ b/tests/test_helpers.rs
@@ -43,7 +43,9 @@ impl BrokerImpl {
 #[derive(Debug)]
 pub struct TestConfig {
     pub bootstrap_brokers: Vec<String>,
+    #[allow(dead_code)]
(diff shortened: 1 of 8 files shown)


tests/test_helpers.rs
```diff
@@ -43,7 +43,9 @@ impl BrokerImpl {
 #[derive(Debug)]
 pub struct TestConfig {
     pub bootstrap_brokers: Vec<String>,
+    #[allow(dead_code)]
```
crepererum, line 46: not actually dead code, just not used by all end2end tests and I didn't wanna add yet another top-level crate just for the test infra.

alamb (commented):
👨‍🍳 👌  -- looks like (another) very nice improvement to me

Thank you @crepererum

alamb (approved):

PR #246, opened 2024-08-06
chore: drop `pin-project-lite`

crepererum:
Actually we don't need this, all our nested types fall roughly into three categories:

- **no perf critical:** Use `Box::pin` to create a new allocation `Pin<Box<T>>`
- **already alloated:** They where already wrapped in a `Box<T>` (e.g. for enums with vastly different variants), so `Pin<Box<T>>` is the same overhead.
- **`Unpin`:** Some upstream types implement `Unpin`, so you can just use `Pin::new(&mut x)`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,6 @@ futures = "0.3"
 integer-encoding = "4"
 lz4 = { version = "1.23", optional = true }
 parking_lot = "0.12"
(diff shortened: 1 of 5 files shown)


alamb (approved):
Looks like easier to follow code and fewer dependencies. 👍  thank you @crepererum

Issue #239, opened 2024-06-20
feature-request: SASL SCRAM Mechanism


Andrew-Chen-Wang:
Hi I'm fairly new to the repo, and I liked the article originally published for this.

I was wondering how difficult it would be to implement the SCRAM SASL mechanisms as I'm fairly new to the repository, but Redpanda cloud requires it.

crepererum:
A while back SASL plain was implemented (#216). The code from that PR probably needs to be extended for SCRAM. PRs are welcome.

crepererum:
Closed by #247.

PR #247, opened 2024-08-10
feat: support SASL SCRAM mechanism

WenyXu:
#239

support SASL SCRAM-SHA-256 and SCRAM-SHA-512

### DIFF
--- a/src/messenger.rs
+++ b/src/messenger.rs
@@ -12,7 +12,11 @@ use std::{
 
 use futures::future::BoxFuture;
 use parking_lot::Mutex;
-use rsasl::{config::SASLConfig, mechname::MechanismNameError, prelude::Mechname};
+use rsasl::{
+    config::SASLConfig,
+    mechname::MechanismNameError,
+    prelude::{Mechname, SessionError},
+};
 use thiserror::Error;
 use tokio::{
     io::{AsyncRead, AsyncWrite, AsyncWriteExt, WriteHalf},
@@ -197,6 +201,9 @@ pub enum SaslError {
 
     #[error("Invalid sasl mechanism: {0}")]
     InvalidSaslMechanism(#[from] MechanismNameError),
+
+    #[error("Sasl session error: {0}")]
+    SaslSessionError(#[from] SessionError),
 
     #[error("unsupported sasl mechanism")]
     UnsupportedSaslMechanism,
@@ -582,7 +589,7 @@ where
             .iter()
             .map(|mech| Mechname::parse(mech.0.as_bytes()).map_err(SaslError::InvalidSaslMechanism))
             .collect::<Result<Vec<_>, SaslError>>()?;
-        debug!("Supported mechanisms {:?}", mechanisms);
+        debug!(?mechanisms, "Supported SASL mechanisms");
         let prefer_mechanism =
             Mechname::parse(mechanism.as_bytes()).map_err(SaslError::InvalidSaslMechanism)?;
         if !mechanisms.contains(&prefer_mechanism) {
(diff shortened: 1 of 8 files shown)


WenyXu:
> I would like to wait for upstream to merge and release this feature. I agree with the general approach of using a dedicated crate instead of implementing all these algorithms within rskafka. 👍

Now, we can push forward.

src/messenger.rs
```diff
@@ -549,6 +557,63 @@ where
...
+
+        let Credentials { username, password } = config.credentials();
+        let config = SASLConfig::with_credentials(None, username, password).unwrap();
+        let sasl = rsasl::prelude::SASLClient::new(config);
+        let raw_mechanisms = resp.mechanisms.0.unwrap_or_default();
+        let mechanisms = raw_mechanisms
+            .iter()
+            .map(|mech| Mechname::parse(mech.0.as_bytes()).map_err(SaslError::InvalidSaslMechanism))
+            .collect::<Result<Vec<_>, SaslError>>()?;
+        debug!("Supported mechanisms {:?}", mechanisms);
```
crepererum, line 585:
```suggestion
        debug!(?mechanisms, "Supported SASL mechanisms {:?}");
```

src/messenger.rs
```diff
@@ -549,6 +557,63 @@ where
...
+        let prefer_mechanism =
+            Mechname::parse(mechanism.as_bytes()).map_err(SaslError::InvalidSaslMechanism)?;
+        if !mechanisms.contains(&prefer_mechanism) {
+            return Err(SaslError::UnsupportedSaslMechanism);
+        }
+        let mut session = sasl
+            .start_suggested(&[prefer_mechanism])
+            .map_err(|_| SaslError::UnsupportedSaslMechanism)?;
+        // let selected_mechanism = session.get_mechname();
+        debug!("Using {:?} for the SASL Mechanism", mechanism);
```
crepererum, line 595:
```suggestion
        debug!(?mechanism, "Using SASL Mechanism");
```

src/messenger.rs
```diff
@@ -549,6 +557,63 @@ where
...
+                .expect("step errored!");
+
+            data = Some(out.into_inner());
+
+            // Returns `true` if step needs to be called again with another batch of data.
+            state.is_running()
+        } {
+            let authentication_response = self.sasl_authentication(data.take().unwrap()).await?;
+            data = Some(authentication_response.auth_bytes.0);
+        }
```
crepererum, line 615:
```suggestion
        // we step through the auth process, starting on our side with NO data received so far
        let mut data_received: Option<Vec<u8>> = None;

        loop {
            let mut to_sent = Cursor::new(Vec::new());
            let state = session
                .step(data_received.as_deref(), &mut to_sent)?;

            if !state.is_running() {
                break;
            }
            
            let authentication_response = self.sasl_authentication(to_sent.into_inner().await?;
            data_received = Some(authentication_response.auth_bytes.0);
        }
```

Some improvements:

- use slightly more readable loop construct
- separate variables for read and write direction (which also eliminates an unwrap)
- replace `except` with proper error handling (you probably need to extend `SaslError`)
- remove the code comment that mentions that you've copied this code from some example :wink:

crepererum (changes requested):
Some nitpicks:

- structured logging (which is esp. helpful if you never yeet through production logs trying to debug why your auth suddenly broke)
- some general recommendations for the the auth stepping

src/messenger.rs
WenyXu: Resolved in [01f1107](https://github.com/influxdata/rskafka/pull/247/commits/01f1107834d6fb3f116e027ae67625e7eb1a9282)

crepererum (approved):

PR #248, opened 2024-10-21
feat: add `timestamp-query` feature to allow timestamp offset query

fMeow:
This commit allow timestamp query for partition offset. As stated in the doc comment of `OffsetAt`, timestamp-based query is inconsistent, but I think it better to let the users make their own choice instead of total removal of this sometimes useful feature. I add a feature gate to  disable timestamp-based query for offset by default, only those actively looking for timestamp-based query would enable this feature.

- [X] I've read the contributing section of the project [CONTRIBUTING.md](https://github.com/influxdata/rskafka/blob/main/CONTRIBUTING.md).
- [X] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed).

### DIFF
--- a/src/client/consumer.rs
+++ b/src/client/consumer.rs
@@ -502,6 +502,9 @@ mod tests {
                 match at {
                     OffsetAt::Earliest => Ok(inner.lock().await.range.0),
                     OffsetAt::Latest => Ok(inner.lock().await.range.1),
+                    OffsetAt::Timestamp(_) => {
+                        unreachable!("timestamp based offset is tested in e2e test")
+                    }
                 }
             })
         }
--- a/src/client/partition.rs
+++ b/src/client/partition.rs
@@ -23,7 +23,7 @@ use crate::{
     throttle::maybe_throttle,
     validation::ExactlyOne,
 };
-use chrono::{LocalResult, TimeZone, Utc};
+use chrono::{DateTime, LocalResult, TimeZone, Utc};
 use std::{
     ops::{ControlFlow, Deref, Range},
     sync::Arc,
@@ -82,11 +82,6 @@ pub enum Compression {
 }
 
 /// Which type of offset should be requested by [`PartitionClient::get_offset`].
-///
-/// # Timestamp-based Queries
-/// In theory the Kafka API would also support querying an offset based on a timestamp, but the behavior seems to be
-/// semi-defined, unintuitive (even within Apache Kafka) and inconsistent between Apache Kafka and Redpanda. So we
-/// decided to NOT expose this option.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum OffsetAt {
     /// Earliest existing record.
@@ -97,6 +92,13 @@ pub enum OffsetAt {
 
     /// The latest existing record.
     Latest,
+
+    /// Timestamp
+    ///
+    /// The Kafka API supports querying an offset based on a timestamp, but the behavior seems to be
+    /// semi-defined, unintuitive (even within Apache Kafka) and inconsistent between Apache Kafka and Redpanda. Mostly it
+    /// offers timestamp based querying with millisecond precision.
+    Timestamp(DateTime<Utc>),
 }
 
 #[derive(Debug)]
@@ -902,6 +904,7 @@ fn build_list_offsets_request(partition: i32, topic: &str, at: OffsetAt) -> List
     let timestamp = match at {
         OffsetAt::Earliest => -2,
         OffsetAt::Latest => -1,
+        OffsetAt::Timestamp(ts) => ts.timestamp_millis(),
     };
 
     ListOffsetsRequest {
--- a/tests/consumer.rs
+++ b/tests/consumer.rs
@@ -2,9 +2,12 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use assert_matches::assert_matches;
+use chrono::{DateTime, TimeZone, Utc};
 use futures::{Stream, StreamExt};
 use tokio::time::timeout;
 
+use rskafka::client::partition::OffsetAt;
+use rskafka::record::Record;
 use rskafka::{
     client::{
         consumer::{StartOffset, StreamConsumer, StreamConsumerBuilder},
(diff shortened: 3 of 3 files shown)


Cargo.toml
```diff
@@ -74,6 +74,7 @@ compression-gzip = ["flate2"]
 compression-lz4 = ["lz4"]
 compression-snappy = ["snap"]
 compression-zstd = ["zstd"]
+timestamp-query = []
```
crepererum, line 77: features are mostly used to control dependencies, e.g. if you don't need compression support, then you don't need to get a fat binary with all this load. You don't add a dependency here, so no feature is required.

src/client/partition.rs
```diff
@@ -97,6 +97,10 @@ pub enum OffsetAt {
 
     /// The latest existing record.
     Latest,
+
+    #[cfg(feature = "timestamp-query")]
+    /// Timestamp
+    Timestamp(i64),
```
crepererum, line 103: High-level APIs should have strong types. `i64` is not a timestamp, it's a number. So the user doesn't know what the semantics of this is (my guess is that it's something like "milliseconds from unix epoch", but you have to check with the Kafka docs to be sure). Since `rskafka` already depends on `chrono`, `DateTime<Utc>` would be a good candidate.

crepererum (changes requested):
It's definitely a feature that we should support, but it needs a bit more thorough API design (see code-attached comment) and a test.

fMeow:
I have change the code for a better API with `DateTime<Utc>`. Would add tests later.

fMeow:
Sorry for the long wait. I have add test to query timestamp offset. @crepererum

tests/consumer.rs
```diff
@@ -436,3 +501,14 @@ where
...
     };
 }
+
+fn record_with_timestamp_milliseconds(key: &[u8], timestamp: DateTime<Utc>) -> Record {
+    Record {
+        key: Some(key.to_vec()),
+        value: Some(b"hello kafka".to_vec()),
+        headers: std::collections::BTreeMap::from([("foo".to_owned(), b"bar".to_vec())]),
+        timestamp: timestamp
+            .with_nanosecond(timestamp.nanosecond() / 1_000_000 * 1_000_000)
```
crepererum, line 511: what does that do?

tests/consumer.rs
```diff
@@ -415,6 +418,68 @@ async fn test_stream_consumer_start_at_latest_empty() {
...
+        .create_topic(&topic, 1, 1, 5_000)
+        .await
+        .unwrap();
+
+    fn now() -> DateTime<Utc> {
+        let now = std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .expect("system time before Unix epoch");
+        DateTime::from_timestamp(now.as_secs() as i64, now.subsec_nanos()).unwrap()
+    }
```
crepererum, line 443: Isn't this the same as [`Utc::now()`](https://docs.rs/chrono/latest/chrono/offset/struct.Utc.html#method.now)?

tests/consumer.rs
```diff
@@ -415,6 +418,68 @@ async fn test_stream_consumer_start_at_latest_empty() {
...
+    );
+    let record_1 = record_with_timestamp_milliseconds(b"x", now());
+    partition_client
+        .produce(vec![record_1.clone()], Compression::NoCompression)
+        .await
+        .unwrap();
+    tokio::time::sleep(Duration::from_millis(100)).await;
+    let ts = now();
+
+    let record_2 = record_with_timestamp_milliseconds(b"y", now());
```
crepererum, line 459:
```suggestion
    let record_2 = record_with_timestamp_milliseconds(b"y", ts);
```

tests/consumer.rs
```diff
@@ -415,6 +418,68 @@ async fn test_stream_consumer_start_at_latest_empty() {
...
+            .partition_client(&topic, 0, UnknownTopicHandling::Retry)
+            .await
+            .unwrap(),
+    );
+    let record_1 = record_with_timestamp_milliseconds(b"x", now());
+    partition_client
+        .produce(vec![record_1.clone()], Compression::NoCompression)
+        .await
+        .unwrap();
+    tokio::time::sleep(Duration::from_millis(100)).await;
```
crepererum, line 456: Since you construct the record with a pre-defined time anyways, do you really have to actually sleep here? Can you not just set `ts` to `record_1.timestamp + Duration::from_millis(100)`?

src/client/partition.rs
```diff
@@ -97,6 +97,9 @@ pub enum OffsetAt {
 
     /// The latest existing record.
     Latest,
+
+    /// Timestamp
```
crepererum, line 101: You may want move the warning about potentially broken behavior from the enum-level (see "Timstamp-based Queries") to this variant.

src/client/consumer.rs
```diff
@@ -502,6 +502,8 @@ mod tests {
                 match at {
                     OffsetAt::Earliest => Ok(inner.lock().await.range.0),
                     OffsetAt::Latest => Ok(inner.lock().await.range.1),
+                    // timestamp based offset is tested in e2e test
+                    OffsetAt::Timestamp(_) => unreachable!(),
```
crepererum, line 506:
```suggestion
                    OffsetAt::Timestamp(_) => unreachable!("timestamp based offset is tested in e2e test"),
```

tests/consumer.rs
fMeow: Tailor timestamp to millisecond. Since the time precision in Kafka is at millisecond level.

tests/consumer.rs
fMeow: Yes. But the now() method is hide behind a feature gate, which is not enabled.

tests/consumer.rs
fMeow: I will give it a try.

fMeow:
@crepererum please have a look

fMeow:
Pls have a look. @crepererum

crepererum:
Ah, last minute issue: could you squash all commits into a single one and name it:

`feat: add `timestamp-query` feature to allow timestamp offset query`

fMeow:
done

crepererum:
Sorry for the CI failure, this will get fixed by #251 (seems to be bad timing).

fMeow:
@crepererum seems good now.

PR #251, opened 2024-11-22
fix: remove unused license

crepererum:
This fails CI.


deny.toml
```diff
@@ -15,7 +15,6 @@ allow = [
     "BSD-2-Clause",
     "BSD-3-Clause",
     "MIT",
-    "Unicode-DFS-2016",
```
alamb, line 18: CI is still failing https://circleci.com/gh/influxdata/rskafka/240564

Maybe this needs to be changed to "Unicode-3.0" rather than deleted

crepererum: fixed

alamb (approved):

PR #253, opened 2025-01-03
feat: Add SASL/OAUTHTOKEN support

### LINKED ISSUE #252: SASL/OAUTHBEARER support
Would you be willing to accept a contribution implementing SASL/OAUTHBEARER support? This should be relatively simple, especially now that the library depends on `rsasl`.

The only wrinkle is that this would probably need to contain some form of callback, since on reconnect the client would likely need to request a new token. This callback should _probably_ be async, since it will usually involve some form of async operation. This means `async_trait` or, if you would prefer to avoid that dependency, requiring users to box their futures themselves with a trait returning `Pin<Box<dyn Future<Output = ...> + 'static>>`.

Thoughts? I would be willing contribute this, but I'd rather nail down the approach you prefer first.

einarmo:
Closes #252.

This adds `SASL/OAUTHBEARER` support to the client.

Since `SaslConfig` no longer always contains `Credentials` I did some minor refactoring and moved the rsasl `SASLConfig` building into a separate method on `SaslConfig`. There is no convenient method for doing Oauth, so we have to implement an rsasl callback for providing a bearer token.

I've chosen the token callback to be asynchronous, since users of this library are likely to use an asynchronous HTTP library, or other OAUTH mechanism, and mixing sync and async is a very bad idea.

One thing I have not done is write any tests, mostly since writing an actual integration test supporting this would be very complicated. Since I have now done some of that myself I have a vague idea of what would be needed:

 - You would need an IDP, we used Azure AD, but you could make that self contained using Keycloak or similar, though configuring that as part of tests is also non-trivial.
 - You would need some form of Oauth plugin, which I believe isn't part of the docker image you use in tests (not entirely sure about that, we use the `strimzi` image).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ tokio = { version = "1.19", default-features = false, features = ["io-util", "ne
 tokio-rustls = { version = "0.26", optional = true, default-features = false, features = ["logging", "ring", "tls12"] }
 tracing = "0.1"
 zstd = { version = "0.13", optional = true }
-rsasl = { version = "2.1", default-features = false, features = ["config_builder", "provider", "plain", "scram-sha-2"]}
+rsasl = { version = "2.1", default-features = false, features = ["config_builder", "provider", "plain", "scram-sha-2", "oauthbearer"]}
 
 [dev-dependencies]
 assert_matches = "1.5"
--- a/src/connection/transport/sasl.rs
+++ b/src/connection/transport/sasl.rs
@@ -1,3 +1,14 @@
+use std::{fmt::Debug, sync::Arc};
+
+use futures::future::BoxFuture;
+use rsasl::{
+    callback::SessionCallback,
+    config::SASLConfig,
+    property::{AuthzId, OAuthBearerKV, OAuthBearerToken},
+};
+
+use crate::messenger::SaslError;
+
 #[derive(Debug, Clone)]
 pub enum SaslConfig {
     /// SASL - PLAIN
@@ -15,6 +26,11 @@ pub enum SaslConfig {
     /// # References
     /// - <https://datatracker.ietf.org/doc/html/draft-melnikov-scram-sha-512-04>
     ScramSha512(Credentials),
+    /// SASL - OAUTHBEARER
+    ///
+    /// # References
+    /// - <https://datatracker.ietf.org/doc/html/rfc7628>
+    Oauthbearer(OauthBearerCredentials),
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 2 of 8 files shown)


einarmo:
Looks like audit is failing due to rust version. A simple fix would be to lock `cargo-deny` to `0.16.2`.

Cargo.toml
```diff
@@ -30,13 +30,14 @@ lz4 = { version = "1.23", optional = true }
 parking_lot = "0.12"
 rand = "0.8"
 rustls = { version = "0.23", optional = true, default-features = false, features = ["logging", "ring", "std", "tls12"] }
+serde = "^1.0.210"
```
crepererum, line 33: This doesn't seem to be used anywhere?

src/protocol/frame.rs
```diff
@@ -163,7 +163,7 @@ mod tests {
 
         data.set_position(0);
         let actual = data.read_message(0).await.unwrap();
-        assert_eq!(actual, vec![]);
+        assert!(actual.is_empty())
```
crepererum, line 166: why was that changed?

src/protocol/frame.rs
```diff
@@ -172,6 +172,6 @@ mod tests {
         client.write_message(&[]).await.unwrap();
 
         let actual = server.read_message(0).await.unwrap();
-        assert_eq!(actual, vec![]);
+        assert!(actual.is_empty())
```
crepererum, line 175: same question as above

src/messenger.rs
```diff
@@ -205,6 +201,12 @@ pub enum SaslError {
     #[error("Sasl session error: {0}")]
     SaslSessionError(#[from] SessionError),
 
+    #[error("Invalid SASL config: {0}")]
+    InvalidConfig(#[from] SASLError),
+
+    #[error("Other SASL error: {0}")]
+    Other(String),
```
crepererum, line 208: I guess you've introduced this variant so that you have a catch-all version for the callback? I'm wondering if the callback should even use `SaslError` in the first place or if it should rather return `Box<dyn std::error::Error + Send + Sync>` and you convert this to a `SaslError` variant like

```rust
#[derive(Error, Debug)]
pub enum SaslError {
    // ...

    #[error("OAuth Callback: {0}")]
    OauthCallback(Box<dyn std::error::Error + Send + Sync>),

    // ...
}
```

src/connection/transport/sasl.rs
```diff
@@ -30,19 +46,100 @@ impl Credentials {
...
+            Self::Plain(credentials)
+            | Self::ScramSha256(credentials)
+            | Self::ScramSha512(credentials) => Ok(SASLConfig::with_credentials(
+                None,
+                credentials.username.clone(),
+                credentials.password.clone(),
+            )?),
+            Self::Oauthbearer(credentials) => {
+                // Fetch the token first, since that's an async call.
+                // The user can and should cache the token as appropriate to their OAUTH provider.
```
crepererum, line 60:
```rust
// The user can and should cache the token as appropriate to their OAUTH provider.
```

Could you move this particular note into the docstring for `OauthCallback`? I think this is important for the user.

crepererum (changes requested):
Looks pretty good for the first round. Couple of smaller nitpicks. I've fixed the CI in #254, so if you rebase this branch, you should be fine.

src/protocol/frame.rs
einarmo: I have no clue why, but on newer rust versions the old approach fails to compile due to type resolution.

Cargo.toml
einarmo: There's a version conflict introduced by strict bounds on `serde` in `rsasl` and `criterion` (I think), this is needed for version resolution.

src/messenger.rs
einarmo: That's probably a good idea. I'll change it :+1:

src/protocol/frame.rs
crepererum: OK, then that's fine :slightly_smiling_face:

src/protocol/frame.rs
crepererum: See https://github.com/influxdata/rskafka/pull/253#discussion_r1901708016

Cargo.toml
crepererum: I'll have a look at that :eyes:

crepererum: With this line removed, I have zero issues. Could you try to remove your `Cargo.lock` and see if it works w/o this line? If not, what exact issue are you facing (i.e. can you post the error message)?

einarmo: Huh. You're right. My bad. I did get some weird error but removing Cargo.lock worked.

einarmo:
Just FYI: I added a second commit, since I ran into a bug in the existing SASL implementation. It seems like `rsasl` sometimes (for PLAIN auth, among others), expects us to send a final message even when `state` is marked as finished.

I changed the code there a bit to account for this, I really don't understand how this could work at all, it should have completely broken PLAIN auth.

crepererum:
> I changed the code there a bit to account for this, I really don't understand how this could work at all, it should have completely broken PLAIN auth.

me neither, but if it works now it works :shrug:

crepererum (approved):
thank you

PR #254, opened 2025-01-03
chore: do not reinstall cargo deny in CI

crepererum:
This is already part of the CI image.


crepererum:
See https://github.com/influxdata/rskafka/pull/253#issuecomment-2568769402 for context.

jdockerty (approved):

PR #255, opened 2025-01-06
fix: SASL/PLAIN

einarmo:
The previous fix did not work. This is now definitely verified to be working and also makes coherent sense.

I made a fix in the OAUTHBEARER PR to fix the SASL flow to send a message even when the final state was `Finished(true)`. My solution instead broke the flow in a _different_ way, making it so that we send the final message twice.

This should  fix it. I didn't catch it before because my kafka configuration was wrong and didn't properly require plain auth, and this particular issue doesn't affect oauthbearer. This time I've verified that it fails with `InvalidSaslState` without this fix, and manages to connect with it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -226,39 +226,49 @@ jobs:
         name: zookeeper
         environment:
           - ALLOW_ANONYMOUS_LOGIN=yes
-      - image: docker.io/bitnami/kafka:3
+      - image: docker.io/bitnami/kafka:3.9.0
         name: kafka-0
         environment:
           - KAFKA_CFG_ZOOKEEPER_CONNECT=zookeeper:2181
           - KAFKA_CFG_BROKER_ID=0
           - ALLOW_PLAINTEXT_LISTENER=yes
-          - KAFKA_CFG_LISTENER_SECURITY_PROTOCOL_MAP=CLIENT:PLAINTEXT,EXTERNAL:PLAINTEXT
-          - KAFKA_CFG_LISTENERS=CLIENT://:9092,EXTERNAL://:9093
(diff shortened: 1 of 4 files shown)


crepererum:
do you think we could extend the end2end tests accordingly? It seems that we already have some tests that try to do that, but it didn't seem to reproduce your issue:

https://github.com/influxdata/rskafka/blob/e65584ddc2caed3b2df83dbc5e33e91260a54be4/tests/client.rs#L28-L50

einarmo:
I'll look into it. My tests were also against the bitnami image, so it should be possible to reproduce the issue there.

einarmo:
Turns out you have to set `KAFKA_SASL_CONNECT` in integration tests.

Also the test just seems to hang before this patch, not entirely sure why. I ran

`TEST_INTEGRATION="1" KAFKA_SASL_CONNECT="localhost:9097" KAFKA_CONNECT="localhost:9097" TEST_BROKER_IMPL="kafka" cargo test test_sasl`

einarmo:
Ah, because of the infinite backoff of course. Better solution is to disable that.

einarmo:
This turned out to be a major hassle to get working.

The CI setup didn't configure any connection that allows SASL, so I added that which produced an error complaining about a missing jaas config. Looking into mounting that I found that doing so is essentially impossible in circleci without major changes to the build. Then I found that you were actually using very old versions of the bitnami image, maybe assuming that `3` pointed to the latest `3.*` version of the image, which isn't actually the case here.

The fix was to update the kafka images and adding some more environment variables, which seems to work.

crepererum (approved):
Awesome. Thanks a lot for going the extra mile and also improve the testing. I really appreciate that!

PR #258, opened 2025-03-04
ci: ensure toolchain is installed

crepererum:


jdockerty (approved):

alamb (approved):
We did the same thing in DataFusion too
- https://github.com/apache/datafusion/pull/14990

👍

alamb:
In case anyone cares the reason this is needed is that `rustup` changed (it no longer automatically installs the toolchain as it did previously)

PR #297, opened 2026-03-14
fix: parse control batch records from Record envelope key

drlukeangel:
## Summary

Fixes parsing of Kafka transactional control batch records. Per the [Kafka protocol spec](https://kafka.apache.org/documentation/#controlbatch), control batch records are wrapped in a standard `Record` envelope — the version and type `Int16` fields are encoded in the Record's **key** field, not directly in the raw byte stream.

The current implementation reads version/type directly from the stream, which causes the `Record` length varint to be misinterpreted as the version field, producing:

```
Unknown control batch record version: 8192
```

This breaks consumption of any topic written by Flink SQL, ksqlDB, or Kafka Streams with `processing.guarantee=exactly_once`, since those frameworks use transactional producers that write commit/abort control batches.

## Changes

- **`ControlBatchRecord::read`**: Now reads a `Record` envelope first, then parses version/type from the Record's key field via a new `from_key()` helper.
- **`ControlBatchRecord::write`**: Now wraps the 4-byte key (version + type) inside a `Record` envelope for correct round-trip serialization.
- **Tests**: Updated to construct control batch records inside Record envelopes using a `wrap_control_key_in_record` helper. Roundtrip test continues to pass.

## Test plan

- [x] `cargo check --lib` passes
- [x] Roundtrip proptest (`test_control_batch_record_roundtrip`) passes
- [x] `test_control_batch_record_unknown_version` passes with envelope-wrapped input
- [x] `test_control_batch_record_unknown_type` passes with envelope-wrapped input
- [x] Verified against Confluent Cloud: successfully consumed messages from Flink SQL transactional output topics (`good-jokes-*`) that previously failed with "Unknown control batch record version: 8192"

## Reproduction

1. Create a Flink SQL job that writes to a Kafka topic (uses transactional producer internally)
2. Attempt to consume from the output topic using rskafka
3. Observe `ReadError::Malformed("Unknown control batch record version: 8192")`
4. With this fix, control batches are correctly parsed and consumption succeeds

### DIFF
--- a/src/protocol/record.rs
+++ b/src/protocol/record.rs
@@ -241,21 +241,22 @@ pub enum ControlBatchRecord {
     Commit,
 }
 
-impl<R> ReadType<R> for ControlBatchRecord
-where
-    R: Read,
-{
-    fn read(reader: &mut R) -> Result<Self, ReadError> {
-        // version
-        let version = Int16::read(reader)?.0;
+impl ControlBatchRecord {
+    /// Parse a `ControlBatchRecord` from a key byte slice.
+    ///
+    /// Per the Kafka spec, the control record's key contains:
+    ///   version: int16 (must be 0)
+    ///   type:    int16 (0 = ABORT, 1 = COMMIT)
+    fn from_key(key: &[u8]) -> Result<Self, ReadError> {
+        let mut cursor = Cursor::new(key);
+        let version = Int16::read(&mut cursor)?.0;
         if version != 0 {
             return Err(ReadError::Malformed(
                 format!("Unknown control batch record version: {version}").into(),
             ));
         }
 
-        // type
-        let t = Int16::read(reader)?.0;
+        let t = Int16::read(&mut cursor)?.0;
         match t {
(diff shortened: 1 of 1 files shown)


src/protocol/record.rs
```diff
@@ -241,21 +241,22 @@ pub enum ControlBatchRecord {
...
+        let version = Int16::read(&mut cursor)?.0;
         if version != 0 {
             return Err(ReadError::Malformed(
                 format!("Unknown control batch record version: {version}").into(),
             ));
         }
 
-        // type
-        let t = Int16::read(reader)?.0;
+        let t = Int16::read(&mut cursor)?.0;
```
crepererum, line 259: Since you're kinda creating a sub-reader here: please check if there are no trailing bytes left at the end, otherwise we shall bail out. That check also requires a new test case.

crepererum (commented):
Good catch. Needs one small adjustment + additional test (see comment) plus `cargo fmt`.

PR #298, opened 2026-06-23
feat: allow selecting rustls crypto provider

Theodus:
Split TLS transport features so downstream users can choose the rustls crypto provider instead of always enabling ring.

Keep the existing `transport-tls` feature as a ring-backed compatibility alias, and add `transport-tls-aws-lc-rs` plus `transport-tls-no-provider` for applications that want `aws-lc-rs` or want to install/select another provider themselves.

Update TLS cfg gates to use the provider-neutral transport feature and make the ignored TLS integration test select an explicit provider so all-features builds do not depend on rustls provider auto-selection.


crepererum (approved):
thank you

PR #300, opened 2026-09-04
fix: include message send in request timeout

WenyXu:
Closes #

Apply the configured request timeout to the complete broker request lifecycle: acquiring the write lock, sending and flushing the Kafka frame, and waiting for the response.

Previously the timeout started only after `send_message()` completed, so a stalled TCP write or flush could block indefinitely. On send timeout, poison the Messenger and let the existing retry path discard the cached broker connection before reconnecting.

Add regression coverage for blocked sends and for request deadlines that include the send phase.

- [ ] I've read the contributing section of the project [CONTRIBUTING.md](https://github.com/influxdata/rskafka/blob/main/CONTRIBUTING.md).
- [ ] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,7 @@ proptest = "1"
 proptest-derive = "0.8"
 rustls-pki-types = "1.11"
 rdkafka = { version = "0.39", default-features = false, features = ["libz", "tokio", "zstd"] }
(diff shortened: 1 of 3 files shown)


crepererum (approved):
makes sense, thank you

